PR #31, opened 2026-06-01
Drive the CLI from an argtree Cli spec (argtree 0.1.2)

JPHutchins:
## What this is

Replaces camas's hand-rolled `argparse` parser with a declarative [`argtree`](https://github.com/JPHutchins/argtree) `Cli` spec, and consumes the **typed** result in dispatch instead of an `Any`-typed `argparse.Namespace`.

Originally opened as an experiment gated on three argtree gaps. **Those shipped in argtree 0.1.2**, so this now uses the real APIs and is ready for review.

## ✅ Blockers resolved in argtree 0.1.2

| Issue | Shipped as | Used here |
|---|---|---|
| JPHutchins/argtree#9 | `build_parser(..., parser_class=)` + public `add_arguments()` | `build_parser(Cli, parser_class=CamasArgumentParser, ...)` — no more `parents=[vanilla]` dance |
| JPHutchins/argtree#10 | documented "keep the spec in an interpreted module" pattern | `parser.py` excluded from `mypycify`; argtree added to `[build-system] requires` |
| JPHutchins/argtree#11 | clear `ConfigError` naming the shadowing field | confirmed the `list` → `list_` rename (flag stays `--list`) |

## What changed

| File | Change |
|---|---|
| `src/camas/main/parser.py` | `Cli` NamedTuple spec (one field per arg); `build_parser` constructs it directly into `CamasArgumentParser` via `parser_class=`; new `reconstruct()` wraps `from_namespace`. Stays interpreted (excluded from mypyc). |
| `src/camas/main/dispatch.py` | Reads a typed `Cli` instead of poking the namespace; resolves the env-dependent `--effects` default here (`None` ⇒ absent ⇒ `default_effects_expr()`, `""` ⇒ list Effects). |
| `setup.py` | `parser.py` added to `_main_excluded` (mypyc). |
| `pyproject.toml` | `argtree>=0.1.2` in runtime deps **and** `[build-system] requires`. |
| `tests/main/test_parser.py` | Read the typed `Cli` via `reconstruct` rather than the raw namespace. |

The declarative spec is the **core**; three runtime/state-dependent behaviors stay imperative *around* it — the dynamic per-matrix-axis `--PY` flags, the `task | expression` positional metavar, and the `--effects` default. argtree gives the real parser back, so these escape hatches remain available; camas just lands as "declarative core + imperative shell."

## Why bother — the payoff

The win isn't the prettier spec; it's **types**. `dispatch` now consumes `Cli` (`expression: str | None`, `dry_run: bool`, …) instead of an `argparse.Namespace` where every field is `Any`. Under five strict type-checkers that removes a class of implicit-`Any` reads.

## Verification (all green)

- `--help` output **byte-identical** to the current parser
- **639 tests pass**, **100% coverage** maintained
- **All 5 type-checkers clean** — mypy, pyright, ty, zuban, pyrefly
- **mypyc wheel builds and runs end-to-end**: compiled `dispatch.so` → interpreted `parser.py`; `build_parser()` returns a `CamasArgumentParser` through `parser_class=`, dynamic `--PY` axis flags work

## Open question for review

Is the typed-dispatch readability worth (a) a runtime dependency on argtree and (b) `parser.py` no longer being mypyc-compiled? Leaning yes; not an obvious slam-dunk over the working argparse.

### DIFF
--- a/src/camas/main/parser.py
+++ b/src/camas/main/parser.py
@@ -139,11 +139,10 @@ def build_parser(state: TasksState = EMPTY_STATE) -> argparse.ArgumentParser:
 	"""Build the CLI argument parser.
 
 	The argument surface is declared once on :class:`Cli`; :func:`argtree.build_parser`
-	turns it into a vanilla ``ArgumentParser``, which a :class:`CamasArgumentParser`
-	adopts via argparse's ``parents=`` so the custom ``--help`` (tasks/Effects/Try
-	listings) layers on top. When ``state`` is :class:`LoadOk` with tasks, known task
-	names appear in the positional metavar so the usage line reads like a list of
-	subcommands.
+	constructs it straight into a :class:`CamasArgumentParser` (via ``parser_class=``)
+	so the custom ``--help`` (tasks/Effects/Try listings) layers on top. When ``state``
+	is :class:`LoadOk` with tasks, known task names appear in the positional metavar so
+	the usage line reads like a list of subcommands.
 
 	>>> from camas.core.task import Task
 	>>> from camas.main.state import LoadOk
@@ -155,12 +154,12 @@ def build_parser(state: TasksState = EMPTY_STATE) -> argparse.ArgumentParser:
 	True
 	"""
 	tasks_for_metavar: Mapping[str, TaskNode] = state.tasks if isinstance(state, LoadOk) else {}
-	vanilla: Final = argtree_build_parser(Cli, add_help=False)
-	parser: Final = CamasArgumentParser(
+	parser: Final = argtree_build_parser(
+		Cli,
+		parser_class=CamasArgumentParser,
 		prog="camas",
 		description=DESCRIPTION,
 		formatter_class=argparse.RawDescriptionHelpFormatter,
-		parents=[vanilla],
 	)
 	parser.state = state
 	metavar: Final = expression_metavar(tasks_for_metavar)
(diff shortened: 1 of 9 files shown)


src/camas/main/parser.py
```diff
@@ -52,6 +55,54 @@ def default_effects_expr() -> str:
...
+	)
+	tree: bool = arg("--tree", help="print every defined task's expanded tree and exit")
+	check: bool = arg("--check", help=describe_check_help())
+	effects: str | None = arg(
+		"--effects",
+		nargs="?",
+		const="",
+		help="tuple of Effect instances; pass with no value to list available Effects "
+		"(default: Termtree, or Status('github') when GITHUB_ACTIONS=true)",
+	)
```
JPHutchins, line 93: This is tricky. Does the | None really stand in for an empty tuple? Anyway, that's not really what happens. The default is TermTree, unless GitHub CI environment variable is set, and then there's the feature of adding configurable defaults.

My point is that there usually is a default, and it can't be known until runtime, and it should be displayed here, correctly, rather than giving the misleading "| None".

JPHutchins: Aw, the "| None" is for the purpose of listing effects. IDK if we should do this, it feels like help has everything that's needed. Is there precedent for a CLI param meaning "show me the possible args" when left empty?

JPHutchins:
Holding off, unsure of value. Maybe useful if sub commands are added.

Issue #32, opened 2026-06-07
PEP 723 for tasks.py


JPHutchins:
Could tasks.py be made standalone for tooling that supports PEP 723? I suppose it would need an if name == main block for entry point.

An option for non-Python projects. Also could opt into locking the camas version.

Possible to support type-checking? Obviously camas --check would still work.

JPHutchins:
## Context

This issue proposes letting `tasks.py` carry [PEP 723](https://peps.python.org/pep-0723/) inline script metadata so it can be run standalone with PEP 723-aware tooling (uv, pipx run, etc.), and asks how that interacts with version pinning and type-checking.

These notes capture decisions from a planning session that resolved adjacent issues (Default discovery in #21, versioning in #20), so the PEP 723 work can be picked up with the surrounding design already settled.

## What PEP 723 actually buys us

A PEP 723 block at the top of `tasks.py` looks like:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["camas>=0.1.8"]
# ///
```

That gives the file two things:
1. A declared Python interpreter floor.
2. A declared `camas` (and any other) dependency, with a pin if desired.

Combined with an `if __name__ == "__main__":` block that dispatches into camas, the file becomes self-bootstrapping under tooling that honors PEP 723 — relevant for non-Python projects that want a single-file build script without committing to a `pyproject.toml` + venv layout.

## How this composes with the rest of the design

Three orthogonal concerns landed in the planning session; PEP 723 only owns one of them.

| Concern | Mechanism |
|---|---|
| Dependency / interpreter declaration | PEP 723 inline metadata (this issue) |
| API stability contract | `camas.v0` namespace (see #20) |
| Runtime behavior (default task/effects) | `default = Default(...)` binding (see #21) |

The import statement `from camas.v0 import Task, Default, ...` still resolves the same way whether or not the file has a PEP 723 header — PEP 723 is a tooling-layer concern, not an import-time one. So PEP 723 adoption does not block or constrain #20 or #21, and vice versa.

## Decided

[...]

Issue #25, opened 2026-05-29
`[tool.camas.tasks]` silently ignores `cwd=` and `help=`


JPHutchins:
Tasks defined in `[tool.camas.tasks]` (pyproject.toml) silently ignore `cwd=` and `help=` keyword arguments instead of applying or rejecting them.

### Why
`src/camas/main/expression.py` `eval_node` builds `kw = {k.arg: k.value for k in keywords ...}` (collecting *all* kwargs) but only reads a subset:
- `Task`: reads `cmd`, `name`, `env` — drops `cwd`, `help`
- `Sequential` / `Parallel`: reads `name`, `matrix`, `env` — drops `cwd`, `help`

So this TOML task silently runs in the wrong directory:
```toml
[tool.camas.tasks]
build = 'Task("cargo build", cwd="src-tauri")'   # cwd ignored → runs in project root
```
and `help=` text never appears in `--list` / `<task> --help`.

The Python `Task`/`Group` constructors accept `cwd`/`help`, so the TOML path is strictly less capable — and the gap is invisible (no error).

### Suggestions
- Thread `cwd`/`help` through `eval_node` for the TOML path (preferred — full parity with `tasks.py`), or
- Raise on unrecognized kwargs, e.g. `Task(...) in [tool.camas.tasks]: unsupported keyword 'cwd'`.

_Found while evaluating camas as a poe replacement — wrote a `[tool.camas.tasks]` version of a project's tasks and noticed `cwd`/`help` vanish._

JPHutchins:
Interacts with #21: a `[tool.camas.defaults]` layer is the natural home for project-wide `cwd`/`help`, so the `eval_node` parity fix here should be designed against that schema rather than landing kwarg plumbing twice. Worth coordinating order — either resolve defaults first and thread the merged kwargs through `eval_node`, or land strict-kwarg rejection now so silent drops surface before #21 expands the surface area.

JPHutchins:
Went with **full parity** (the issue body's preferred option) in #38 rather than strict-kwarg rejection.

[...]

Issue #21, opened 2026-05-29
Default


JPHutchins:
Default task node, default effect, default GitHub task node, etc.

JPHutchins:
Leaning more towards "config", of which defaults would be one field, for example.

JPHutchins:
## Planning notes carried forward

This comment captures the design discussion so a future implementer can pick up #21 without re-deriving it. The follow-up suggestion to widen this to a "config" concept is acknowledged at the end; the body below describes the narrow `Default` shape that was designed in detail.

## Scope decided for the first PR

A single new public type, `Default`, declared in `tasks.py`:

```python
default = Default(task=..., effects=...)
```

Bare `camas` invokes `default.task`. `default.effects` becomes the new fallback for `--effects` when the user doesn't pass one. Both fields are optional.

A `Version` type was also explored for the same PR and rejected — see #20 for the reasoning (the short version: `from camas.v0 import ...` already encodes the API-version contract via Python's import system, so a runtime `Version` declaration mostly duplicates the namespace signal). The `Default` work below is independent of that decision and can land on its own.

## API shape

```python
class Default(NamedTuple):
    task: TaskNode | None = None
    effects: tuple[Effect[Any], ...] | None = None
```

Lives in `src/camas/core/meta.py`, re-exported through `camas.v0` (and therefore the top-level `camas` package).

`task` is `TaskNode`-only, not `TaskNode | str`. This matches the strongly-typed CLI: users bind task variables and compose them with Python expressions, never by string lookup. The string-by-name alternative was considered (it would have been convenient for a future TOML loader, see #25) and rejected to keep the API consistent — there is one way to refer to a task, and it's the binding.

## Loader changes

- New helper `name_scope_default(scope)` in the loader: reads `scope["default"]`, type-checks `isinstance(val, Default)`. A wrong type raises rather than being silently ignored, so typos surface.
- Extend `LoadOk` with `default: Default | None = None` (appended at the end so existing constructions stay valid). Known call sites to touch: `state.EMPTY_STATE`, `dispatch.run_cli`, `dispatch._load_py`, and the test fixture in `test_dispatch.py`.

## Dispatch wiring (`src/camas/main/dispatch.py`)

Two branches need the new field:

1. **Default task selection.** When `args.expression is None and default.task is not None`, set `resolved = default.task` and fall through to the existing post-resolve pipeline (matrix overrides, passthrough, dry-run, `run()`). No new code path beyond the resolve step.

[...]

Issue #20, opened 2026-05-29
Versioning scheme


JPHutchins:

JPHutchins:
# Versioning scheme

Working notes from a prior planning session. Captures what was decided, what was explored and rejected, and where the work lands in the tree.

## Decision

**The import path is the versioning mechanism.** `from camas.v0 import ...` is the contract. There is no `Version` type, no `version = Version(...)` declaration in `tasks.py`, and no runtime version-check warning.

Rationale: Python's import system already statically resolves `from camas.v0 import Task, Sequential, Parallel, Default`. If `camas.v0` exists, the user is on a camas that ships the v0 namespace. If it does not, they get an `ImportError` naming the package. A bespoke `Version` NamedTuple compared against `importlib.metadata.version("camas")` would mostly duplicate that signal while adding API surface we would later have to deprecate.

Starting namespace is `v0` (not `v1`) because camas is currently 0.x. The mapping is `Version(N, ...) <=> camas.vN` if `Version` is ever reintroduced.

## What was explored and rejected

A `Version` NamedTuple in `tasks.py`:

```python
class Version(NamedTuple):
    major: int
    minor: int
    patch: int
    label: str | None = None  # "rc1", ".dev123", "+abc123" — informational only
```

Semantics that were worked out before the type was dropped:

- Comparison is tuple-only on `(major, minor, patch)`. `label` is stored, displayed, never compared.
- Severity: warn to stderr and continue, not hard error. Easy to upgrade, doesn't break CI on bump.
- Discovery via `name_scope_version(scope)` with `isinstance` check; wrong type raises so typos surface.
- `check_version()` compares against `importlib.metadata.version("camas")`. Unparseable installed strings short-circuit so a dev install never trips a false alarm.

Why it was dropped: the only thing `Version` added beyond the namespace was minor/patch granularity (e.g. "feature added in 0.1.8, installed is 0.1.5"). For camas's small, slow-moving surface that case surfaces as `ImportError` or `AttributeError` on the specific symbol the `tasks.py` tries to use, which is no worse than a custom warning and arrives at exactly the line that needs to be fixed.

It is easier to add `Version` back later when a concrete motivating case shows up than to deprecate it now.

## The v0 namespace (what actually lands)

Create `src/camas/v0/__init__.py` as the canonical public surface. Flat re-export list. This file **is** the v0 contract: any change to its exported names or their signatures is breaking and forces a new `camas.v1` namespace.

Contents:

[...]

PR #34, opened 2026-06-10
feat(api): add the camas.v0 versioned namespace

JPHutchins:
## Summary

Implements the decision recorded in #20: **the import path is the versioning mechanism.** `camas.v0` is the public API, versioned the way semver versions the package: `v0` pairs with camas 0.x and is exactly as loose as semver says 0.x is — the surface prefers to grow, breaking changes remain possible until 1.0 (deliberate and release-noted, never accidental). At 1.0 the contract hardens: a stable-era namespace never removes or changes an exported name, a breaking change forces the next `camas.vN`, and old namespaces keep shipping, so files written against one keep working across upgrades.

- **`src/camas/v0/__init__.py`** — the contract module: `Task`, `Sequential`, `Parallel`, `TaskNode`, `Effect`, the `TaskEvent` stream types, the `LeafState` types, and the `Completion` outcomes. Doctests demo the definers and pattern-matching on `Finished | Skipped`.
- **Top-level `camas`** keeps exactly its four existing re-exports, now sourced from `.v0` — `from camas import Task` is unchanged and simply tracks the latest generation.
- **`tests/test_v0.py`** makes changes deliberate rather than accidental: pins the exact export list (an unintended addition/removal/re-binding fails CI and forces an edit of `V0_CONTRACT`) and asserts re-export identity against the `camas.core.*` definitions.
- **Fixture migration**: `tests/fixtures/effect-plugin/` (the out-of-tree effect exemplar, symlinked from `examples/`) and the `Loud` doctest in the package docstring move from `camas.core.*` imports to `camas.v0`.
- **README**: new Versioning section stating the semver-zero semantics, the top-level relationship, and that feature-level pinning belongs to the dependency declaration (pyproject / PEP 723), not the import path.

## Deviations from the #20 notes

1. **`Default` is not in the inaugural surface.** The notes' implementation order bundled #21, but its shape is still contested ("leaning more towards config") and the namespace PR stays small without it. Under semver-zero looseness it can land in v0 whenever it settles. The #21 notes bless this decoupling ("can land on its own").
2. **`Completion`/`Finished`/`Skipped` added to the export list.** Exported types carry them in public fields (`CompletedEvent.completion`, `Completed.completion`), and any effect that distinguishes pass/fail must pattern-match on them — omitting them would force plugin authors back into `camas.core.completion`. This is the only gap in the field-type closure of the notes' list.
3. **v0 is semver-zero loose, not frozen from birth** (per review discussion). The notes' "any change forces v1" rule would have burned `v1` on the first alpha iteration while the package was still 0.x, breaking the `vN` ↔ semver-major mapping the same notes establish. The freeze activates at 1.0 instead.

Out of scope, per the notes: `camas.v0.effects.*`, deprecation warnings on internals, `run`/`run_cli` promotion (#32's open question), TOML config (#25).

## Test plan

- [x] `camas '{format_check, lint, mypy, pyright, ty, zuban, test}'` green locally (pyrefly excluded only because the sandbox worktree path makes it match zero files; CI covers it — and passed on the first push)
- [x] `camas coverage` holds the 100% gate
- [x] `tests/test_v0.py` pins the contract; effect-plugin fixture test exercises the migrated imports end-to-end via runpy


JPHutchins:
Per review discussion: v0 is semver-zero loose, not frozen from birth — the surface prefers to grow, breaking changes stay possible until 1.0 (deliberate and release-noted), and the hard freeze activates at 1.0. This keeps the `vN` ↔ semver-major mapping coherent instead of burning `v1` on the first alpha iteration. Reworded in 9267e2c (module docstrings, README, test docstring); the contract-pinning test itself is unchanged — its job was always deliberateness, not permanence.

JPHutchins (commented):
I feels like everything is backwards. Like the concrete interfaces should be defined in the vN files and `camas.__init__` should always re-export the latest v.

src/camas/v0/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+from .leaf_state import Waiting as Waiting
+from .task import Group as Group
+from .task import Parallel as Parallel
+from .task import Sequential as Sequential
+from .task import Task as Task
+from .task import TaskNode as TaskNode
+from .task_event import CompletedEvent as CompletedEvent
+from .task_event import OutputEvent as OutputEvent
+from .task_event import StartedEvent as StartedEvent
+from .task_event import TaskEvent as TaskEvent
```
JPHutchins: This should probably only export the same that the root does. All namespaces must be kept 1:1.

README.md
```diff
@@ -137,6 +137,14 @@ The downside is that fork PRs get a read-only `GITHUB_TOKEN` from GitHub and can
...
+
+The top-level `camas` namespace re-exports `camas.v0` **1:1** — the two always expose the identical set of names (the definers plus the full plugin contract: `TaskNode`, the `TaskEvent` stream, `LeafState`, `Completion`). `from camas import ...` is the unversioned alias that tracks the latest generation; `from camas.v0 import ...` pins it — reach for the pinned form in anything you distribute or won't revisit. Everything else (`camas.core.*`, `camas.main.*`) is internal: importable today, but with no stability promise.
+
+To pin a minimum camas *feature* level, use your dependency declaration (`camas>=0.x` in `pyproject.toml`, or PEP 723 inline metadata) — the import path covers API shape; the package pin covers feature availability.
+
```
JPHutchins, line 147: It wouldn't ALWAYS be in v0. This is way to specific and needs to be restated.

src/camas/__init__.py
```diff
@@ -22,6 +22,14 @@
 uses ``help=`` only in a handful of places), but available for
 cryptic commands.
 
+The full API — these definers plus the plugin contract (events,
+states, completions) — is defined in :mod:`camas.v0` and re-exported
+here 1:1, so ``camas`` and ``camas.v0`` always expose the identical
+set of names. ``from camas import ...`` is the unversioned alias that
+tracks the latest generation; ``from camas.v0 import ...`` pins the
+generation. v0 is semver-zero loose while camas is 0.x and hardens
+into a frozen contract at 1.0.
```
JPHutchins: Not sure that this comment has merit.

src/camas/__init__.py
```diff
@@ -285,11 +291,24 @@
...
+from .v0 import OutputEvent as OutputEvent
+from .v0 import Parallel as Parallel
+from .v0 import Running as Running
+from .v0 import Sequential as Sequential
+from .v0 import Skipped as Skipped
+from .v0 import StartedEvent as StartedEvent
+from .v0 import Task as Task
+from .v0 import TaskEvent as TaskEvent
+from .v0 import TaskNode as TaskNode
+from .v0 import Waiting as Waiting
```
JPHutchins: This should only export what it did before, preserving the existing namespace.

src/camas/core/__init__.py
```diff
@@ -1,3 +1,5 @@
 # SPDX-License-Identifier: MIT
 # SPDX-FileCopyrightText: 2026 JP Hutchins
-"""Engine: the task-tree AST, matrix expansion, async execution, and the Effect protocol."""
+"""Engine: matrix expansion, async execution, traversal, and rendering over the
+:mod:`camas.v0` types.
+"""
```
JPHutchins, line 5: Insane to refer to v0 here, utter rubbish.

JPHutchins (commented):
References to v0 are made throughout that are guaranteed to be WRONG in the future. Namespace was polluted. This is off the rails, investigate and resolve.

src/camas/__init__.py
```diff
@@ -235,10 +241,11 @@
...
 	...         from collections.abc import Sequence
-	...         from camas import Effect, Parallel, Task
-	...         from camas.core.leaf_state import LeafState
-	...         from camas.core.task import TaskNode
-	...         from camas.core.task_event import OutputEvent, TaskEvent
+	...         from camas import Parallel, Task
+	...         from camas.v0 import Effect
+	...         from camas.v0.leaf_state import LeafState
+	...         from camas.v0.task import TaskNode
+	...         from camas.v0.task_event import OutputEvent, TaskEvent
```
JPHutchins: Shouldn't these import from root? Or if we're demonstrating pinning, be consistent and don't only pin a few?

src/camas/v0/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+	...         case Finished(returncode=rc):
+	...             return f"exited {rc}"
+	...         case Skipped(returncode=rc):
+	...             return f"skipped (prior rc={rc})"
+	>>> verdict(Finished(0, 0.5, (b"ok",)))
+	'exited 0'
+	>>> verdict(Skipped(1))
+	'skipped (prior rc=1)'
+
+See ``examples/effect-plugin/`` for a complete out-of-tree effect.
```
JPHutchins: Reduce to a single line, this is LLM slop.

tests/test_v0.py
```diff
@@ -0,0 +1,101 @@
...
+# SPDX-FileCopyrightText: 2026 JP Hutchins
+
+"""Pin the public-API surface.
+
+The top-level ``camas`` namespace and the current versioned namespace
+(``camas.v0``) expose the identical four headline definers, kept 1:1; the
+rest of the public surface for writing effects lives in that generation's
+submodules. The version namespaces are semver-zero loose until 1.0 — they
+may change, but only on purpose, so an accidental addition, removal, or
+re-binding of the headline surface fails here.
```
JPHutchins: Delete

JPHutchins (commented):
Good progress, just a bit of slop.

src/camas/v0/__init__.py
JPHutchins: Done in 06f9336 — `camas.v0` now exports exactly the four headline definers (`Effect`, `Parallel`, `Sequential`, `Task`), identical to root. `tests/test_v0.py` pins `public_names(camas) == public_names(camas.v0)` so they can't drift.

src/camas/__init__.py
JPHutchins: Rewritten and trimmed (06f9336, b0f2c31): the note is now three lines and version-neutral — it just says these four are the unversioned alias for the latest generation and points to the README.

src/camas/__init__.py
JPHutchins: Reverted in 06f9336 — root re-exports exactly the original four (`Effect`, `Parallel`, `Sequential`, `Task`).

src/camas/__init__.py
JPHutchins: Made consistent in b0f2c31 — the doctest (and the effect-plugin fixture) now pin every name to `camas.v0` submodules. Importing all from root isn't possible since the contract types (`TaskNode`, the event/state types) aren't on the root surface, so the demo pins uniformly rather than mixing root and pinned imports.

src/camas/v0/__init__.py
JPHutchins: Reduced to a single line in b0f2c31.

tests/test_v0.py
JPHutchins: Deleted in b0f2c31 (test modules are exempt from the docstring rules).

src/camas/__init__.py
```diff
@@ -285,11 +289,11 @@
...
-Effect protocol details: :class:`camas.core.effect.Effect`. Per-task
+Effect protocol: subclass ``Effect`` (see ``examples/effect-plugin/``). Per-task
 help: ``camas <task> --help``.
 """
 
-from .core.effect import Effect as Effect
-from .core.task import Parallel as Parallel
-from .core.task import Sequential as Sequential
-from .core.task import Task as Task
+from .v0 import Effect as Effect
```
JPHutchins, line 296: **[code-review · altitude · future reference]** The headline re-export is hand-maintained in three places that must stay in lockstep: this block, `camas/v0/__init__.py`'s four exports, and `tests/test_v0.py`'s `HEADLINE` + `PUBLIC_API`. `test_v0.py` pins `public_names(camas) == public_names(camas.v0)`, so the 1:1 is enforced — but nothing enforces that `camas` tracks the *latest* generation. When `camas.v1` lands and this should flip to `from .v1 import ...`, a missed edit leaves `from camas import …` silently pinned to v0 while the README says it's the latest, and the 1:1 test still passes (root == v0). A single source of truth (e.g. driving the alias from the newest version package) or a test asserting root tracks the highest `camas.vN` would close that gap. No action needed for v0.

src/camas/core/effect.py
```diff
@@ -1,43 +1,14 @@
...
+from typing import TypeAlias
 
-from .task_event import TaskEvent
-
-if TYPE_CHECKING:
-	from .leaf_state import LeafState
-	from .task import TaskNode
+from ..v0.task_event import TaskEvent
 
 EventSink: TypeAlias = Callable[[int, TaskEvent], Awaitable[None]]
```
JPHutchins, line 13: **[code-review · simplification · future reference]** Now that the `Effect` protocol lives in `camas/v0/effect.py`, this module is ~14 lines of header/scaffolding around a single one-line `TypeAlias`, and its only consumer is `core/execution.py` (which already imports `TaskEvent` from `..v0.task_event` directly). Could fold `EventSink` into the top of `execution.py` and drop the module. Navigation/scaffolding cost only.

src/camas/core/task.py
```diff
@@ -1,17 +1,14 @@
...
-from pathlib import Path
-from types import MappingProxyType
 from typing import TYPE_CHECKING, NamedTuple, TypeAlias
 
 if TYPE_CHECKING:
-	from collections.abc import Mapping
+	from ..v0.task import Task
 
 
 class VarBinding(NamedTuple):
```
JPHutchins, line 14: **[code-review · placement · future reference]** After the split this module holds only `task_label` plus the matrix tuples `VarBinding`/`MatrixBinding`, whose sole consumer is `core/matrix.py` (it imports all three from here). The name now mirrors `v0/task.py` but defines no task type. Folding `VarBinding`/`MatrixBinding` into `matrix.py` would sit them next to their only user and shrink this to the `task_label` accessor. Pre-existing placement the split made more visible; no behavior impact.

JPHutchins:
## Code review (automated, max effort)

Reviewed the full PR diff across five angles (line-by-line correctness, removed-behavior, cross-file/import-hazard, Python pitfalls/type-identity, and tests/doctests). **No correctness issues** — the move is behavior-preserving: the relocated public types are semantically identical to their originals, type identity is intact (so `match`/`isinstance`/`issubclass` and effect discovery still target the same objects), no runtime use slipped under `TYPE_CHECKING`, `camas.v0` has no dependency on the engine, and coverage holds at 100%.

Posted three **cleanup/altitude** notes inline for future reference (none blocking): the v1 alias-divergence gap, the now-tiny `core/effect.py`, and the matrix-type placement in `core/task.py`.

PR #37, opened 2026-06-12
fix(dispatch): accept hyphenated aliases for underscore task names

### LINKED ISSUE #24: Hyphenated task name (e.g. `camas test-all`) fails with cryptic `unsupported syntax: BinOp(...)`
Invoking a task whose name contains a hyphen fails confusingly: the CLI arg is parsed as a Python expression (`test - all` → a subtraction `BinOp`) instead of being looked up or suggested.

### Repro
`tasks.py`:
```python
from camas import Parallel, Task
test_all = Parallel(Task("echo hi"))
```
```
$ camas test-all
error: unsupported syntax: BinOp(left=Name(id='test', ctx=Load()), op=Sub(), right=Name(id='all', ctx=Load()))
```
(`camas test_all` works fine.)

### Why
`dispatch_arg` (`src/camas/main/dispatch.py`) does `if arg in tasks: return tasks[arg]`, otherwise parses `arg` as an expression. A hyphenated name isn't a key (Python `tasks.py` vars are underscore-named) so it parses as a `BinOp` and falls through to the `unsupported syntax` branch in `expression.py:eval_node`.

This bites because every peer runner (just, Task, poe, npm scripts) uses hyphens, so `test-all` is the natural thing to type. TOML keys *can* be hyphenated and resolve fine, which sharpens the inconsistency.

### Suggestions
1. Normalize on lookup: try `arg`, then `arg.replace("-", "_")` before treating `arg` as an expression.
2. Failing that, when an expression resolves to a bare `Name`/`BinOp` that isn't a known task, emit a friendly error like `no task named 'test-all'; did you mean 'test_all'?` (you already do the "did you mean" style for unknown `Name`s — extend it to the hyphen/`BinOp` case).

_Found while evaluating camas as a drop-in replacement for poethepoet._

JPHutchins:
## Problem

`camas test-all` for a `tasks.py` binding named `test_all` failed with a cryptic error:

```
$ camas test-all
error: unsupported syntax: BinOp(left=Name(id='test', ctx=Load()), op=Sub(), right=Name(id='all', ctx=Load()))
```

The arg was parsed as a Python expression (`test - all`, a subtraction) and fell through to the `unsupported syntax` branch. Every peer runner (just, Task, poe, npm scripts) uses hyphens, so `test-all` is the natural thing to type. (#24)

## Fix

Per the issue discussion (underscore is canonical, but allow the `-` fallback users assume by convention), `dispatch_arg`:

1. matches the name **verbatim** first — so a hyphenated `[tool.camas.tasks]` key (`matrix-ci`) still resolves directly;
2. then matches with **hyphens folded to underscores** — so `camas test-all` reaches the `test_all` binding;
3. otherwise, if the arg is identifier-shaped (a `_NAME_LIKE` bare name, possibly hyphenated, with no parens/quotes/braces/operators), emits the friendly `no task named 'test-all' (known: ...)` error instead of the `BinOp` dump — extending the existing bare-`Name` handling to the hyphen case;
4. only genuine expressions (carrying parens/quotes/braces/operators) take the inline-expression path, unchanged.

This also lets the implementation drop the `ast.parse` round-trip in `dispatch_arg` in favor of a single regex.

## Tests

- `test_dispatch_arg_hyphen_folds_to_underscore_binding` — `dispatch_arg("test-all", {"test_all": ...})` resolves.
- `test_dispatch_arg_hyphenated_unknown_errors` — friendly error, not a `BinOp` dump.
- `test_hyphenated_invocation_of_underscore_binding` — end-to-end `camas test-all` against a `tasks.py` defining `test_all`.
- Existing `test_dispatch_arg_hyphenated_name` (literal hyphenated TOML key) still passes.

Full suite: 532 passed. ruff/mypy/pyright/ty clean.

Fixes #24

### DIFF
--- a/src/camas/main/dispatch.py
+++ b/src/camas/main/dispatch.py
@@ -5,8 +5,8 @@
 
 from __future__ import annotations
 
-import ast
 import asyncio
+import re
 import sys
 from pathlib import Path
 from typing import TYPE_CHECKING, Final
@@ -41,18 +41,26 @@ if TYPE_CHECKING:
 	from ..v0.task import TaskNode
 
 
+_NAME_LIKE: Final = re.compile(r"^[A-Za-z_][A-Za-z0-9_-]*$")
+"""A bare task name — possibly hyphenated — as opposed to a camas expression
+(which carries parens, quotes, braces, or operators)."""
+
+
 def dispatch_arg(arg: str, tasks: Mapping[str, TaskNode]) -> TaskNode:
-	"""Interpret a CLI arg: task name (possibly hyphenated) ⇒ lookup, else inline expression."""
-	if arg in tasks:
-		return tasks[arg]
-	try:
-		parsed = ast.parse(arg, mode="eval")
-	except SyntaxError:
-		return parse_expression(arg, tasks=tasks)
-	if isinstance(parsed.body, ast.Name):
-		name = parsed.body.id
+	"""Interpret a CLI arg: task name ⇒ lookup, else inline expression.
+
+	The name is matched verbatim first (so a hyphenated ``[tool.camas.tasks]``
+	key resolves directly), then with hyphens folded to underscores — so a
+	``tasks.py`` binding ``test_all`` is reachable as the conventional
+	``camas test-all`` even though ``test-all`` is a subtraction, not a name.
+	Underscore is canonical; the hyphen spelling is the convention-driven alias.
+	"""
+	for candidate in (arg, arg.replace("-", "_")):
+		if candidate in tasks:
+			return tasks[candidate]
+	if _NAME_LIKE.match(arg):
 		known = ", ".join(sorted(tasks)) or "none"
-		print(f"error: no task named {name!r} (known: {known})", file=sys.stderr)
+		print(f"error: no task named {arg!r} (known: {known})", file=sys.stderr)
 		sys.exit(2)
 	return parse_expression(arg, tasks=tasks)
 
--- a/src/camas/main/dispatch.py
+++ b/src/camas/main/dispatch.py
@@ -42,18 +42,18 @@ if TYPE_CHECKING:
 
 
 _NAME_LIKE: Final = re.compile(r"^[A-Za-z_][A-Za-z0-9_-]*$")
-"""A bare task name — possibly hyphenated — as opposed to a camas expression
-(which carries parens, quotes, braces, or operators)."""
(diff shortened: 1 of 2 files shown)


src/camas/main/dispatch.py
```diff
@@ -41,18 +41,26 @@
...
-		return parse_expression(arg, tasks=tasks)
-	if isinstance(parsed.body, ast.Name):
-		name = parsed.body.id
+	"""Interpret a CLI arg: task name ⇒ lookup, else inline expression.
+
+	The name is matched verbatim first (so a hyphenated ``[tool.camas.tasks]``
+	key resolves directly), then with hyphens folded to underscores — so a
+	``tasks.py`` binding ``test_all`` is reachable as the conventional
+	``camas test-all`` even though ``test-all`` is a subtraction, not a name.
+	Underscore is canonical; the hyphen spelling is the convention-driven alias.
```
JPHutchins: Comment may be stating what the code does, should be removed.

JPHutchins (commented):
Mostly OK, has some LLM slop docs.

src/camas/main/dispatch.py
```diff
@@ -41,18 +41,26 @@
 	from ..v0.task import TaskNode
 
 
+_NAME_LIKE: Final = re.compile(r"^[A-Za-z_][A-Za-z0-9_-]*$")
+"""A bare task name — possibly hyphenated — as opposed to a camas expression
+(which carries parens, quotes, braces, or operators)."""
```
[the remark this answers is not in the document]
JPHutchins: Good catch — reworded. `_NAME_LIKE`'s docstring no longer lists "operators" as an expression marker (which contradicted allowing `-`); it now says a `-` is taken as part of the name (the convention alias), never as subtraction, and that expressions are distinguished by parens/quotes/braces.

src/camas/main/dispatch.py
JPHutchins: Removed — trimmed the trailing "Underscore is canonical…" line; the preceding sentence already conveys the verbatim-then-fold behavior and the rationale.

JPHutchins:
**Review status** — both threads resolved in `cde479a`:

- ✅ `_NAME_LIKE` "operators" wording (Copilot) → reworded; the `-` is documented as part of the name (the convention alias), not a subtraction operator.
- ✅ trailing "Underscore is canonical…" docstring line → removed as redundant.

CI green. Nothing pending — ready for re-review.

PR #38, opened 2026-06-12
fix(toml): thread cwd/help through the expression evaluator

### LINKED ISSUE #25: `[tool.camas.tasks]` silently ignores `cwd=` and `help=`

JPHutchins:
## Problem

Tasks defined in `[tool.camas.tasks]` (and the inline expression mini-language) silently ignored `cwd=` and `help=`. `eval_node` (`src/camas/main/expression.py`) built `kw = {...}` collecting *all* kwargs but only read a subset:

- `Task`: read `cmd`, `name`, `env` — dropped `cwd`, `help`
- `Sequential` / `Parallel`: read `name`, `matrix`, `env` — dropped `cwd`, `help`

So this ran in the wrong directory, silently:

```toml
[tool.camas.tasks]
build = 'Task("cargo build", cwd="src-tauri")'   # cwd ignored → ran in project root
```

and `help=` never reached `--list` / `<task> --help` — even though the Python `Task`/`Group` constructors accept both. The TOML path was strictly less capable, invisibly. (#25)

## Fix

`eval_node` now threads `cwd` and `help` into `Task` / `Sequential` / `Parallel`, giving the TOML path **full parity** with `tasks.py` (the issue's preferred option). `assign_key_name` and `resolve_refs` already preserve both, so they survive TOML-key naming and ref resolution.

Per the issue discussion, this is the parity fix; a future project-wide defaults layer (#21) can build on it rather than re-plumbing `eval_node`.

## Tests

- `test_load_tasks_threads_cwd_and_help` — the exact issue scenario end-to-end (`Task("cargo build", cwd="src-tauri", help=...)` and a `Parallel(..., cwd=, help=)`) through `load_tasks`.
- `test_parse_task_value_threads_cwd_and_help` — `parse_task_value` unit.
- `test_parse_expression_threads_cwd_and_help` — inline expression mini-language.

(Comparisons use `Path` equality, not repr, so they hold on Windows too.) Full suite: 532 passed. ruff/mypy/ty clean.

Fixes #25

### DIFF
--- a/src/camas/main/expression.py
+++ b/src/camas/main/expression.py
@@ -202,6 +202,8 @@ def eval_node(
 						cmd=eval_cmd(cmd_node),
 						name=eval_opt_str(kw.get("name")),
 						env=eval_env(kw.get("env")),
+						cwd=eval_opt_str(kw.get("cwd")),
+						help=eval_opt_str(kw.get("help")),
 					)
 				case "Sequential" | "Parallel":
 					ctor = Sequential if name == "Sequential" else Parallel
@@ -210,6 +212,8 @@ def eval_node(
 						name=eval_opt_str(kw.get("name")),
 						matrix=eval_matrix(kw.get("matrix")),
 						env=eval_env(kw.get("env")),
+						cwd=eval_opt_str(kw.get("cwd")),
+						help=eval_opt_str(kw.get("help")),
 					)
 				case "Ref":
 					ref_name_node = args[0] if args else kw.get("name")
--- a/tests/main/test_expression.py
+++ b/tests/main/test_expression.py
@@ -3,10 +3,14 @@
 
 from __future__ import annotations
 
+import inspect
+from pathlib import Path
+
 import pytest
 
 from camas import Parallel, Sequential, Task
 from camas.main.expression import parse_expression
+from camas.v0.task import Group
 
 
 @pytest.mark.parametrize(
@@ -135,6 +139,28 @@ def test_parse_expression_threads_cwd_and_help() -> None:
 	assert parse_expression('Parallel(Task("a"), cwd="work", help="grp")') == Parallel(
 		Task("a"), cwd="work", help="grp"
 	)
+
+
+def test_eval_node_threads_every_public_constructor_kwarg() -> None:
+	"""Drift guard: a new public kwarg on Task/Sequential/Parallel fails here until
+	it's both wired through ``eval_node`` and given a sentinel below — so a silently
+	dropped kwarg (the #25 bug) can't reappear unnoticed."""
+	# fragment + (attribute, expected value) for each constructor kwarg.
+	samples = {
+		"name": ("name='n'", "name", "n"),
+		"env": ("env={'K': 'v'}", "env", {"K": "v"}),
+		"cwd": ("cwd='d'", "cwd", Path("d")),
+		"help": ("help='h'", "help", "h"),
+		"matrix": ("matrix={'X': ('1',)}", "matrix", {"X": ("1",)}),
+	}
+	for cls, prefix, variadic in (
+		(Task, "Task('c', ", "cmd"),
+		(Group, "Parallel(Task('a'), ", "tasks"),
+	):
+		for kw in set(inspect.signature(cls).parameters) - {variadic}:
+			frag, attr, expected = samples[kw]  # KeyError flags an unsampled new kwarg
+			node = parse_expression(prefix + frag + ")")
+			assert getattr(node, attr) == expected, (cls.__name__, kw)
 
 
 # Parser-side fluent syntax: strings inside literals and the README one-liner.
--- a/tests/main/test_expression.py
+++ b/tests/main/test_expression.py
@@ -126,6 +126,17 @@ def test_parse_explicit_none_name() -> None:
 	assert parse_expression('Task("hi", name=None)') == Task("hi", name=None)
 
 
+def test_parse_expression_threads_cwd_and_help() -> None:
+	"""``cwd`` / ``help`` reach the constructors from the expression mini-language,
+	at parity with ``tasks.py`` — not silently dropped (#25)."""
+	assert parse_expression('Task("cargo build", cwd="src-tauri", help="Build")') == Task(
+		"cargo build", cwd="src-tauri", help="Build"
+	)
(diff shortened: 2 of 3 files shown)


src/camas/main/expression.py
```diff
@@ -210,6 +212,8 @@ def eval_node(
 						name=eval_opt_str(kw.get("name")),
 						matrix=eval_matrix(kw.get("matrix")),
 						env=eval_env(kw.get("env")),
+						cwd=eval_opt_str(kw.get("cwd")),
+						help=eval_opt_str(kw.get("help")),
```
JPHutchins, line 216: Anyway to flag when the interface adds args so we don't forget to keep this mapping up to date? Dynamic?

JPHutchins (commented):
Great, a question about maintenance burden though.

src/camas/main/expression.py
JPHutchins: Two ways to make this fail loudly when the constructor surface grows:

1. **Signature-introspection guard test.** Assert that the kwargs `eval_node` constructs from match `inspect.signature(Task)` / `Group` exactly (minus `cmd`/`*tasks`). Add a public param to `Task` without wiring `eval_node` → the test goes red. Test-only, no runtime change, keeps this PR parity-only.

2. **Dynamic dispatch + reject-unknown in `eval_node`.** Map each `k.arg` through a per-name `eval_*` table and raise on any key not in it. This also closes the adjacent silent-drop gaps Copilot flagged (typos like `cwdd=` / `matirx=`, and `**`-unpacking at the call level). Heavier, and it crosses into the "strict rejection" alternative we deliberately deferred for this PR (scope here was parity).

My lean: (1) in this PR — it locks the mapping to the signature without changing behavior — and track (2) as a focused follow-up, since reject-unknown is a behavior change touching the same code the #21 defaults layer will revisit. Happy to add the guard test here, or do the full dynamic+reject in one pass if you'd rather. Your call.

src/camas/main/expression.py
```diff
@@ -202,6 +202,8 @@ def eval_node(
 						cmd=eval_cmd(cmd_node),
 						name=eval_opt_str(kw.get("name")),
 						env=eval_env(kw.get("env")),
+						cwd=eval_opt_str(kw.get("cwd")),
+						help=eval_opt_str(kw.get("help")),
 					)
```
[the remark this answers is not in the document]
JPHutchins: Valid — at the `Task(...)` call level `eval_node` still silently drops both `**`-unpacking (`k.arg is None`) and unknown/typo kwargs like `cwdd=`. (The per-field `eval_env` / `eval_matrix` do reject `**` *within* their dicts, but not at the constructor call.) This is the "strict rejection" alternative from #25, deliberately deferred for this parity-only PR per the issue discussion. I've raised the maintenance/sync angle with the maintainer in the thread on line 216 and proposed tracking reject-unknown-kwargs as a focused follow-up rather than widening scope here.

src/camas/main/expression.py
```diff
@@ -210,6 +212,8 @@ def eval_node(
 						name=eval_opt_str(kw.get("name")),
 						matrix=eval_matrix(kw.get("matrix")),
 						env=eval_env(kw.get("env")),
+						cwd=eval_opt_str(kw.get("cwd")),
+						help=eval_opt_str(kw.get("help")),
 					)
```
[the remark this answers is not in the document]
JPHutchins: Same class of gap as the `Task` case — `Sequential` / `Parallel` (and `Ref`) silently ignore unknown/typo kwargs (`matirx=`, `cwdd=`). Folding reject-unknown into all three is the right fix; tracking it as a follow-up alongside the maintainer discussion on line 216 so this PR stays parity-only.

JPHutchins:
**Review status** — the parity change itself is green; the open threads are scope decisions for you, no code changed yet:

- ⏳ **Your Q (line 216 — keep the kwarg mapping in sync):** I proposed two paths and held for your pick:
  1. a signature-introspection **guard test** (parity-only, goes red if `Task`/`Group` gains a public param `eval_node` doesn't thread) — my lean; or
  2. **dynamic dispatch + reject-unknown** in `eval_node` (also closes the Copilot points below, but it's the "strict rejection" path we deferred per the #25 discussion).
  Happy to add (1) here or do (2) in one pass — just say which.
- 📌 **Copilot (lines 207 / 217):** unknown/typo kwargs (`cwdd=`, `matirx=`) and `**`-unpacking are still silently dropped at the call level. That's the strict-reject alternative, deferred for this parity-only PR; tracked under your Q above.

So: **not blocked** to merge as parity-only, or hold for a decision on (1)/(2).

src/camas/main/expression.py
JPHutchins: Yes 1. is approved

JPHutchins: Done (afa072f) — added a signature-introspection guard: it round-trips every public `Task`/`Group` kwarg through `parse_expression` and asserts it survives, with the param list driven off `inspect.signature`. A new constructor kwarg that isn't wired through `eval_node` (or lacks a sentinel in the test) fails the suite. Strict reject-unknown-kwargs (option 2) remains tracked separately.

PR #41, opened 2026-06-12
refactor(demo): single source for typed and executed command

JPHutchins:
Follow-up to #36 (already merged), addressing the review comment on `.github/scripts/demo-payload.sh`: the keystrokes were typed from `CMD` but the executed command was hard-coded (`exec ... camas all`), so the two could drift if the demo command ever changed.

Make `CMD` a bash array (`CMD=(camas all)`): the keystrokes are typed from `"${CMD[*]}"` and the command is run as `exec env -u GITHUB_ACTIONS "${CMD[@]}"`, so the displayed and executed argv share one source.

(This commit was made on the #36 branch just after it merged, so it missed the merge — re-landing it here.)

- `shellcheck` clean
- `shfmt -d` no diff

### DIFF
--- a/.github/scripts/demo-payload.sh
+++ b/.github/scripts/demo-payload.sh
@@ -6,13 +6,16 @@
 set -euo pipefail
 
 PROMPT='\033[1;32m❯\033[0m '
-CMD='camas all'
+CMD=(camas all)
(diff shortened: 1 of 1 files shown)


JPHutchins:
Dumb AF

PR #45, opened 2026-06-13
feat(cli): add --init to scaffold a starter tasks.py

### LINKED ISSUE #18: --init
Creates a commented tasks.py demonstrating many features with infallible cross platform dummy tasks.

JPHutchins:
Closes #18

`camas --init` writes a commented starter `tasks.py` into the current directory and exits. The template demonstrates, with infallible cross-platform placeholder commands (`python -c ...`):

- leaf `Task` (tuple cmd as-is, string cmd shlex-split, bare-string coercion inside groups)
- `Sequential` (ordered, short-circuit) / `Parallel` (concurrent) composition
- a matrix with `{AXIS}` interpolation, scoped `env=`, and `help=`
- a `Config` default task (#21), so bare `camas` runs green immediately after scaffolding
- the PEP 723 header + `run_cli(globals())` standalone block, per the canonical shape settled in #32

Behavior:
- refuses to overwrite an existing `tasks.py` (stderr + exit 2)
- handled in both dispatch arms, so a broken tasks file elsewhere up the tree (`LoadErr`) doesn't block scaffolding a fresh one
- `init` added to `RESERVED_FLAGS` so a matrix axis named `INIT` can't collide

Tests cover the write/refuse paths, both dispatch arms, loader round-trip (task names + promoted `Config.default_task`), an end-to-end bare-`camas` run of the scaffold (proving the placeholders are infallible), `--init`-then-`--list` via the real CLI, and the standalone `python tasks.py --list` path.

Verified locally: ruff format/check, mypy, pyright, ty, zuban, full suite at 100% coverage (pyrefly + unscoped variants via CI).

### DIFF
--- a/src/camas/main/init.py
+++ b/src/camas/main/init.py
@@ -6,72 +6,26 @@
 from __future__ import annotations
 
 import sys
-from typing import TYPE_CHECKING, Final
+from pathlib import Path
+from typing import Final
 
-if TYPE_CHECKING:
-	from pathlib import Path
-
-STARTER: Final = '''\
-# /// script
-# requires-python = ">=3.10"
-# dependencies = ["camas"]
-# ///
-"""Project tasks — run with ``camas``, or standalone via ``uv run tasks.py``.
-
-Every task below is a cross-platform placeholder (``python -c ...``);
-replace the placeholders with your real commands:
-
-	lint = Task("ruff check .")
-	test = Task("cargo test", cwd=Path("rust"))
-	build = Task("npm run build")
-"""
-
-from camas import Config, Parallel, Sequential, Task, run_cli
-
-# A leaf is any shell command. Tuple form runs as-is; string form is
-# shlex-split. A bare string inside Sequential/Parallel coerces to an
-# anonymous Task.
-hello = Task(("python", "-c", "print('hello from camas')"))
-
-# matrix= clones the subtree per axis value, interpolating {NAME} into
-# cmd/env/cwd; env= is scoped to the subtree. Pin an axis from the CLI:
-# camas greet --NAME Grace
-greet = Parallel(
-	Task(("python", "-c", "import os; print(os.environ['GREETING'] + ', {NAME}!')")),
-	matrix={"NAME": ("Ada", "Grace")},
-	env={"GREETING": "hello"},
-	help="say hello to everyone at once",
-)
-
-# Sequential runs in order, short-circuiting on the first failure;
-# Parallel runs concurrently. They nest freely.
-ci = Sequential(
-	hello,
-	Parallel(greet, "python --version"),
-	name="ci",
-)
-
-# Discovered by type (the binding name never matters): bare `camas` runs
-# default_task — or github_task under GitHub Actions, falling back to
-# default_task when unset.
-_ = Config(default_task=ci)
-
-# Optional: with the PEP 723 header above, any PEP 723-aware tool runs
-# this file directly (`uv run tasks.py --list`). camas auto-discovery
-# ignores this block.
-if __name__ == "__main__":
-	run_cli(globals())
-'''
+STARTER: Final = (Path(__file__).parent / "starter.py").read_text(encoding="utf-8")
+"""The packaged :mod:`camas.main.starter` template, scaffolded verbatim.
+mypyc ships the original ``.py`` next to the compiled artifact, so the read
+works identically from a compiled wheel."""
 
 
 def write_starter_tasks_py(directory: Path) -> int:
 	"""Write :data:`STARTER` to ``directory/tasks.py`` — the exit code for
-	``camas --init``. Refuses to overwrite an existing file.
+	``camas --init``. Exclusive create: an existing file is never touched, and
+	any ``OSError`` (exists, permissions, ...) becomes a clean error exit.
 	"""
 	target: Final = directory / "tasks.py"
-	if target.exists():
-		print(f"error: {target} already exists", file=sys.stderr)
+	try:
+		with target.open("x", encoding="utf-8") as f:
+			f.write(STARTER)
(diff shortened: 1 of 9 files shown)


src/camas/main/init.py
```diff
@@ -0,0 +1,77 @@
...
+	test = Task("cargo test", cwd=Path("rust"))
+	build = Task("npm run build")
+"""
+
+from camas import Config, Parallel, Sequential, Task, run_cli
+
+# A leaf is any shell command. Tuple form runs as-is; string form is
+# shlex-split. A bare string inside Sequential/Parallel coerces to an
+# anonymous Task.
+hello = Task(("python", "-c", "print('hello from camas')"))
```
JPHutchins: Shouldn't this be "python -c print("hello from camas`)"? And all the others should be like that too?

src/camas/main/init.py
```diff
@@ -0,0 +1,77 @@
...
+
+if TYPE_CHECKING:
+	from pathlib import Path
+
+STARTER: Final = '''\
+# /// script
+# requires-python = ">=3.10"
+# dependencies = ["camas"]
+# ///
+"""Project tasks — run with ``camas``, or standalone via ``uv run tasks.py``.
```
JPHutchins: Specifiy that PEP723 is optional and explain what it does and how to use it.

src/camas/main/init.py
```diff
@@ -0,0 +1,77 @@
...
+
+from __future__ import annotations
+
+import sys
+from typing import TYPE_CHECKING, Final
+
+if TYPE_CHECKING:
+	from pathlib import Path
+
+STARTER: Final = '''\
```
JPHutchins: TBH should it just be a file?

JPHutchins (commented):
Some misuse and other ideas.

src/camas/main/init.py
JPHutchins: Done in e7bbf2b — the template is now a real packaged module, `src/camas/main/starter.py` (excluded from mypyc compilation; mypyc ships the original .py next to the .so, so `init.py` reads it back as text identically from a compiled wheel). It now gets formatted, linted, and type-checked by all six checkers like any other module, and pytest's doctest collection imports it so it sits under the coverage gate.

src/camas/main/init.py
JPHutchins: Done in e7bbf2b — the template docstring now explains that the `# /// script` block is optional PEP 723 inline metadata, that with the `__main__` block it makes the file standalone under PEP 723-aware tools (`uv run tasks.py`, `pipx run tasks.py`), that it's where camas gets version-pinned, and that both are inert under plain `camas` and deletable.

src/camas/main/init.py
JPHutchins: Done in e7bbf2b — all placeholders are string-form now. One nuance: the -c payload must stay double-quoted (`Task("python -c \"print('hello from camas')\"")`), because string cmds are shlex-split — a bare `python -c print('hello from camas')` has its single quotes consumed by shlex mid-token and the payload arrives as `print(hello from camas)`, a SyntaxError. The comment above `hello` teaches exactly this (quote the payload; tuple form skips shlex when quoting gets hairy).

This change also surfaced the Windows CI failure as a real bug: standalone `python tasks.py --list` output (the `×` in the matrix annotation) went through cp1252 because `run_cli` lacked the UTF-8 stdio reconfigure that `entrypoint.main` has. Both entry points now share `dispatch.reconfigure_stdio_utf8`.

src/camas/main/init.py
```diff
@@ -0,0 +1,77 @@
...
+	"""Write :data:`STARTER` to ``directory/tasks.py`` — the exit code for
+	``camas --init``. Refuses to overwrite an existing file.
+	"""
+	target: Final = directory / "tasks.py"
+	if target.exists():
+		print(f"error: {target} already exists", file=sys.stderr)
+		return 2
+	target.write_text(STARTER, encoding="utf-8")
+	print(f"Wrote {target}\n\nTry:\n  camas --list\n  camas greet --help\n  camas")
+	return 0
```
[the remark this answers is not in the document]
JPHutchins: Done in e7bbf2b — switched to exclusive create: `target.open("x")` inside a single `except OSError` that prints `error: ...` and returns 2. `FileExistsError` is an `OSError`, so the race is gone and permission errors get the same clean exit instead of a traceback.

JPHutchins:
The nix failures were a third compiled-context gotcha: nixpkgs' mypyc doesn't define `__file__` during compiled module init, so the module-level template read crashed with `KeyError: '__file__'` (cibuildwheel's newer mypyc populates it, which is why the wheels passed). `STARTER` is now a call-time `starter_text()` — same pattern `mypyc.py` already uses. Verified locally: `nix build .#checks.x86_64-linux.package-with-check` passes (592 tests in the sandbox against the compiled package), full camas gate green.

src/camas/main/starter.py
```diff
@@ -0,0 +1,54 @@
...
+	name="ci",
+)
+
+# Discovered by type (the binding name never matters): bare `camas` runs
+# default_task — or github_task under GitHub Actions, falling back to
+# default_task when unset.
+_ = Config(default_task=ci)
+
+if __name__ == "__main__":
+	run_cli(globals())
```
JPHutchins, line 57: Needs a comment about why this would be necessary - it's the PEP 723 flow

JPHutchins (commented):
Pretty good

src/camas/main/starter.py
JPHutchins: Done — restored a comment at the block pointing at the PEP 723 flow (running the file directly dispatches through camas; inert under discovery), with the full explanation staying in the docstring.

Issue #58, opened 2026-06-19
camas mcp server serves stale code after the camas package itself changes (self-dev / editable installs)


JPHutchins:
## Problem

The `camas mcp` stdio server is a long-lived process. It already re-resolves `tasks.py` when that file's mtime changes (`Session.refresh` in `mcp/serve.py`), so newly added/renamed tasks show up via `camas_list`/`camas_run` without a restart.

What it does **not** detect is a change to the **camas package code itself**. When you develop camas using its own MCP server (editable / `uv run camas mcp`), edits to camas source are invisible to the running server: it keeps the old `Task`, `wire.RunRequest`, tool input schemas, handlers, etc. that were imported at process start.

In the worst case it can't even *load* a `tasks.py` that uses newly added public API, so every tool call fails until the client restarts the server.

## Repro

1. Start the server (`.mcp.json`: `uv run camas mcp --rich`).
2. On a feature branch, add a new public `Task` kwarg (e.g. `mutates`) and use it in the repo's own `tasks.py`:
   ```python
   format = Task("uv run ruff format .", mutates=True)
   ```
3. Call any camas MCP tool. `camas_list` returns:
   ```
   Tasks unavailable — .../tasks.py failed to evaluate
   (TypeError: Task.__init__() got an unexpected keyword argument 'mutates').
   ```
   `camas_run`'s advertised input schema is also stale (no new fields, old required-ness).

The task *definitions* refreshed from disk, but the **code** evaluating them is the version loaded when the server booted.

## Impact

Only bites when the project under development **is** camas (or when camas is an editable dep being hacked on). Normal users pin camas as a fixed dependency, so the server's code is stable and only task definitions change — already handled. So this is a contributor/dogfooding papercut, not a user-facing correctness bug.

## Possible resolutions (pick one)

- **Docs note**: "after editing camas itself, restart the MCP server" — cheapest; this is normal for any server hosting its own code.
- **Clearer error hint**: when a load fails with an `unexpected keyword argument` `TypeError`, append a hint suggesting the server may be running stale camas code and to restart it.
- **Detect + re-exec**: watch the installed camas package mtime (alongside the tasks-source mtime) and `os.execv` the server to reload on change. Heaviest; a long-lived Python process can't safely hot-reload its own imported modules in place.

Suggest the docs note + error hint; re-exec is probably overkill.

## Found while

[...]

PR #59, opened 2026-06-19
feat: `camas --under=<duration>` — time-budgeted run mode (#54)

### LINKED ISSUE #54: feat: `camas --under=<duration>` — run everything that fits a time budget in parallel
## Motivation

camas now records per-task duration estimates (the timings cache, surfaced via `camas_list`). With estimates in hand, camas could offer a **time-budgeted run mode**: pack as much useful work as possible into a wall-clock budget, rather than making the user (or an AI agent) hand-pick which tasks are "quick enough" for a tight inner loop.

## Proposed UX

```
camas --under=1s
```

Run, in parallel, every task whose estimate fits the budget — maximizing the checks you get for a given wall-clock cost. Great for the agent inner loop ("run whatever validates in ~1s before I move on") and for humans who don't want to remember which task is fast.

Open questions:
- Packing strategy: maximize task count? coverage/importance? greedy by estimate?
- Tasks with no estimate yet — include optimistically, or exclude until timed?
- Is the budget per-task (each leaf must be `< budget`) or whole-run (the parallel group's max must be `< budget`)?

## Wrinkle: mutating tasks

Some tasks mutate the workspace — formatters and lint-fixers (`ruff format`, `ruff check --fix`, `cargo clippy --fix`, …). They can't be tossed into an arbitrary parallel group: they race with read-only checks over the same files, and order matters (format *then* check).

So camas needs to **know a task mutates**. Proposal: a "mark" (e.g. `mutates=True` on `Task`, or a dedicated marker) the budget scheduler can use to:

1. **(preferred)** Run marked/mutating tasks **in sequence, first**, then launch the parallel read-only group. The agent never burns a turn hand-formatting something that could have been automatic.
2. (rejected) Exclude mutating tasks from `--under` runs entirely — bad: the agent loses the auto-fix benefit and may waste turns doing it manually.

## Notes

- Split out of the MCP PR (`feature/#19/mcp`) to keep that PR focused. This is a general camas scheduling feature, independent of MCP — though `camas_run` could later expose a budget arg too.
- Depends on the timings cache (landed on the MCP branch).

JPHutchins:
Closes #54.

Run, in parallel, every leaf whose timed estimate fits a wall-clock budget — the fast inner-loop subset, picked for you instead of by hand.

## What it adds

- **`Task(mutates=True)`** — marks a leaf that writes the workspace (a formatter / auto-fixer). Threaded through every Task reconstruction site (matrix expand/specialize, passthrough, naming, expression eval/render) and round-trips through `to_expression`/`parse_expression`.
- **CLI `camas --under=<duration>`** — `1s`, `500ms`, `2m`, `1h`, or a bare number of seconds. Budgets the named task/expression, or the `Config` default when none is given.
- **MCP `camas_run` `under` argument** — same budget for agents (the primary use case, per #54). `task` becomes optional when `under` is set (budgets the project default). The response carries a structured `budget` report of what was selected and excluded.

## Scheduling

Selected leaves are run as `Sequential(*mutating, Parallel(*read_only))` — the mutating leaves first, in tree order, then the read-only rest in parallel, so a formatter never races a checker over the same files (the issue's preferred resolution). Sequential's short-circuit semantics are preserved (a failing fixer blocks the read-only group, same as a hand-written `fix` step).

```
$ camas --under=1s --dry-run
Time budget 1.00s — selected 6 leaf(s), excluded 2 (2 over budget, 0 untimed).
  over budget: pyright ~4.57s, coverage ~20.98s
lint_fix → format → (mypy | ty | zuban | pyrefly)
```

## Open-question resolutions (from the issue)

- **Packing**: a filter, not a knapsack — every leaf whose estimate fits is selected ("every task whose estimate fits the budget").
- **Budget semantics**: **per-leaf** — a leaf is selected when its own estimate ≤ budget, so the parallel group's wall-clock stays near the budget.
- **Untimed leaves**: **excluded** (strict) — a budget can't bound an un-estimated leaf; the count is reported so it's visible, and a normal run times them. *(your call)*
- **Scope**: the `Config` default task (override with a name/expression); `camas_run` mirrors it (omit `task`). *(your call)*

Leaves are de-duplicated by value before scheduling, so a leaf shared across the tree runs once.

## Tests & dogfooding

- New `tests/core/test_budget.py`; new `--under`/`parse_duration` cases in the parser & dispatch tests; new `camas_run under=` cases in the MCP tests. **100% coverage**, all five type checkers, lint, and format clean — verified by running the project's own `camas check` / `camas coverage`.
- `tasks.py` marks its `format` / `lint_fix` leaves `mutates=True`, so camas dogfoods the feature.

## Notes

- Filed #58 while working on this: the long-lived `camas mcp` server runs stale camas code after the camas package itself changes, so it couldn't load a `tasks.py` using the new `mutates=` kwarg — a self-development papercut. Verification this session went through the `camas` CLI (fresh working-tree code, same task definitions) instead.

### DIFF
--- a/src/camas/core/budget.py
+++ b/src/camas/core/budget.py
@@ -1,0 +1,115 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: 2026 JP Hutchins
+
+"""Time-budgeted scheduling: select the leaves of a task that fit a wall-clock budget."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, NamedTuple, TypeAlias
+
+from ..v0.task import Parallel, Sequential, Task
+from .matrix import expand_matrix
+from .timings import estimate
+from .traversal import flatten_leaves
+
+if TYPE_CHECKING:
+	from collections.abc import Mapping
+
+	from ..v0.task import TaskNode
+	from .timings import TaskLabel, TaskTiming
+
+
+class Fits(NamedTuple):
+	"""A leaf whose estimate is within budget — selected to run."""
+
+	task: Task
+	estimated_s: float
+
+
+class OverBudget(NamedTuple):
+	"""A leaf whose estimate exceeds the budget — excluded."""
+
+	task: Task
+	estimated_s: float
+
+
+class Untimed(NamedTuple):
+	"""A leaf with no recorded estimate — excluded from a strict budget."""
+
+	task: Task
+
+
+Disposition: TypeAlias = Fits | OverBudget | Untimed
+
+
+class BudgetPlan(NamedTuple):
+	"""A budget's verdict over a source task's de-duplicated leaves.
+
+	``node`` is the runnable schedule — mutating leaves in a leading ``Sequential``
+	followed by the read-only remainder as a single ``Parallel`` — or ``None`` when
+	nothing fits.
+	"""
+
+	budget_s: float
+	node: TaskNode | None
+	fits: tuple[Fits, ...]
+	over_budget: tuple[OverBudget, ...]
+	untimed: tuple[Untimed, ...]
+
+
+def classify(task: Task, budget_s: float, timings: Mapping[TaskLabel, TaskTiming]) -> Disposition:
+	"""A leaf's disposition under ``budget_s``, read from its observed estimate.
+
+	>>> from camas.core.timings import TaskTiming
+	>>> classify(Task("a"), 1.0, {"a": TaskTiming(0.5, 1)})
+	Fits(task=Task(cmd='a', name=None, env={}, cwd=None), estimated_s=0.5)
+	>>> classify(Task("a"), 1.0, {"a": TaskTiming(2.0, 1)})
+	OverBudget(task=Task(cmd='a', name=None, env={}, cwd=None), estimated_s=2.0)
+	>>> classify(Task("a"), 1.0, {})
+	Untimed(task=Task(cmd='a', name=None, env={}, cwd=None))
+	"""
+	est = estimate(task, timings)
+	if est is None:
+		return Untimed(task)
+	if est.elapsed_s <= budget_s:
+		return Fits(task, est.elapsed_s)
+	return OverBudget(task, est.elapsed_s)
+
+
+def plan_under(
+	node: TaskNode, budget_s: float, timings: Mapping[TaskLabel, TaskTiming]
+) -> BudgetPlan:
+	"""Partition ``node``'s expanded, de-duplicated leaves by ``budget_s`` and schedule
+	those that fit. Mutating leaves run sequentially first, then the read-only leaves as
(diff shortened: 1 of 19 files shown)


src/camas/core/budget.py
```diff
@@ -0,0 +1,115 @@
...
+
+Disposition: TypeAlias = Fits | OverBudget | Untimed
+
+
+class BudgetPlan(NamedTuple):
+	"""A budget's verdict over a source task's de-duplicated leaves.
+
+	``node`` is the runnable schedule — mutating leaves in a leading ``Sequential``
+	followed by the read-only remainder as a single ``Parallel`` — or ``None`` when
+	nothing fits.
```
JPHutchins: Slop

src/camas/core/budget.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def plan_under(
+	node: TaskNode, budget_s: float, timings: Mapping[TaskLabel, TaskTiming]
+) -> BudgetPlan:
+	"""Partition ``node``'s expanded, de-duplicated leaves by ``budget_s`` and schedule
+	those that fit. Mutating leaves run sequentially first, then the read-only leaves as
+	one parallel group; leaves with no estimate are excluded (a strict budget cannot
+	bound them).
+	"""
```
JPHutchins: Slop

src/camas/core/budget.py
```diff
@@ -0,0 +1,115 @@
...
+	fits = tuple(d for d in dispositions if isinstance(d, Fits))
+	over_budget = tuple(d for d in dispositions if isinstance(d, OverBudget))
+	untimed = tuple(d for d in dispositions if isinstance(d, Untimed))
+	return BudgetPlan(budget_s, schedule(tuple(f.task for f in fits)), fits, over_budget, untimed)
+
+
+def schedule(fitting: tuple[Task, ...]) -> TaskNode | None:
+	"""Order the fitting leaves: mutating leaves sequentially first (formatters before
+	checkers, never racing the read-only group over the same files), then the read-only
+	leaves as one parallel group. ``None`` when nothing fits.
```
JPHutchins: Slop

src/camas/mcp/serve.py
```diff
@@ -366,53 +372,171 @@ async def run_call(session: Session, arguments: dict[str, Any]) -> types.CallToo
...
 
 	Raises:
-		ValueError: when ``req.task`` names no task, an override targets an unknown
-			matrix axis, or passthrough args are applied to a non-leaf task.
+		ValueError: when ``req.task`` is omitted (only ``under`` may omit it), names no
+			task, an override targets an unknown matrix axis, or passthrough args are
+			applied to a non-leaf task.
 	"""
-	if req.task not in tasks:
+	name = req.task
```
JPHutchins: Don't alias

JPHutchins (commented):
Pretty good, some slop

PR #65, opened 2026-06-20
fix(mcp): report missing mcp extra cleanly

### LINKED ISSUE #60: camas mcp crashes with bare ModuleNotFoundError when the [mcp] extra isn't installed (VSCode: "Connection closed")
## Summary

`camas mcp` is dispatched unconditionally, but the server module imports the optional `mcp` package at top level. When the `camas[mcp]` extra is **not** installed, the server process dies immediately with a bare `ModuleNotFoundError`. To an MCP client (VSCode) this surfaces as the opaque `MCP error -32000: Connection closed`, and from the terminal it's an uncaught traceback rather than an actionable "install the extra" message.

The user's instinct is right: if the `mcp` extra is missing, `camas mcp` either shouldn't pretend to work, or should fail with a clear install hint — not crash.

## Environment

- camas `0.1.13` (also reproduces on current `main` @ `9776589`)
- Python 3.14
- Installed **without** the `mcp` extra (i.e. `mcp` package absent from the venv)
- Client: VSCode MCP integration

## Steps to reproduce

1. Install camas without the `mcp` extra (e.g. plain `camas`, no `camas[mcp]`).
2. Configure it as an MCP server (`camas mcp`) in VSCode, **or** run `camas mcp` directly.

## Actual behavior

VSCode:

```
MCP error -32000: Connection closed
```

Terminal:

```
$ uv run camas mcp
Traceback (most recent call last):
  File ".../bin/camas", line 10, in <module>
    sys.exit(main())
  File "src/camas/main/entrypoint.py", line 19, in main
  File ".../site-packages/camas/mcp/cli.py", line 35, in main
    from .serve import serve_stdio
  File ".../site-packages/camas/mcp/serve.py", line 20, in <module>
    from mcp import types
ModuleNotFoundError: No module named 'mcp'
```

## Root cause

`mcp` is an optional extra:

```toml
# pyproject.toml
[project.optional-dependencies]
mcp = ["mcp>=1.24,<2"]
```

But the `mcp` subcommand is always routed, regardless of whether that extra is installed:

```python
# src/camas/main/entrypoint.py
if argv and argv[0] == "mcp":
    importlib.import_module("camas.mcp.cli").main(argv[1:])
    return
```

…and `cli.main()` reaches the server via a plain import:

```python
# src/camas/mcp/cli.py
from .serve import serve_stdio
serve_stdio(argv)
```

…whose module imports the optional dependency at top level, so the import itself explodes:

```python
# src/camas/mcp/serve.py
from mcp import types
from mcp.server.lowlevel import Server
from mcp.server.lowlevel.server import NotificationOptions
```

When `mcp` isn't installed, `from .serve import serve_stdio` raises `ModuleNotFoundError`, which propagates uncaught.

## Expected behavior

`camas mcp` (and `camas mcp init`, `-h`, etc.) should degrade gracefully when the extra is missing — exit non-zero with a clear, actionable message naming the extra, e.g.:

```
camas mcp: requires the optional `mcp` extra. Install it with: pip install "camas[mcp]"  (or uv add "camas[mcp]")
```

`camas mcp -h` / `camas mcp init` arguably shouldn't require `mcp` to be importable at all — `init` only scaffolds `.mcp.json` and help just prints text.

## Suggested fix

The repo already solves exactly this for the `httpx` / `github_checks` extra with a lazy import + friendly error:

```python
# src/camas/effect/github_checks.py
def require_httpx() -> ModuleType:
    try:
        import httpx
    except ImportError as e:
        raise RuntimeError("GitHubChecks: requires feature camas[github_checks]") from e
    return httpx
```

Apply the same shape to the MCP path — e.g. guard the `from .serve import serve_stdio` in `cli.main()` (or add a `require_mcp()` preflight) so a missing `mcp` extra produces the install hint and a clean non-zero exit instead of a traceback / `Connection closed`.

jstar0:
## Summary

Fixes #60.

`camas mcp` currently reaches the server import path before the optional `mcp` dependency is installed. When users install `camas` without the `mcp` extra, the command fails with a bare `ModuleNotFoundError`, which can surface to MCP clients as an opaque connection close.

## Changes

- Catch the missing optional `mcp` dependency at the `camas mcp` serve boundary and exit with a concise extra-focused message: `camas mcp: requires feature camas[mcp]`.
- Re-raise other `ModuleNotFoundError` cases so unrelated import bugs are not hidden.
- Add a regression test that simulates a cold `camas.mcp.serve` import without `mcp` installed.

## Verification

```bash
uv run pytest tests/mcp/test_lazy_import.py -q
uv run pytest tests/mcp/test_lazy_import.py tests/mcp/test_scaffold.py::test_entrypoint_mcp_init_routes_to_scaffold tests/mcp/test_scaffold.py::test_mcp_cli_help_prints_usage -q
uv run ruff check src/camas/mcp/cli.py tests/mcp/test_lazy_import.py
uv run ruff format --check src/camas/mcp/cli.py tests/mcp/test_lazy_import.py
uv run pyright src/camas/mcp/cli.py tests/mcp/test_lazy_import.py
uv run mypy src/camas/mcp/cli.py tests/mcp/test_lazy_import.py
git diff --check
```

### DIFF
--- a/src/camas/mcp/cli.py
+++ b/src/camas/mcp/cli.py
@@ -24,7 +24,11 @@ Options:
 
 
 def main(argv: list[str]) -> None:
-	"""Route ``camas mcp [-h|init|--rich|...]``; ``init`` exits with its own code."""
+	"""Route ``camas mcp [-h|init|--rich|...]``; ``init`` exits with its own code.
+
+	Raises:
+		ModuleNotFoundError: if a dependency other than the optional ``mcp`` extra is missing.
+	"""
 	if "-h" in argv or "--help" in argv:
 		print(HELP)
 		return
@@ -32,6 +36,12 @@ def main(argv: list[str]) -> None:
 		from .scaffold import write_mcp_json
 
 		sys.exit(write_mcp_json(argv[1:]))
-	from .serve import serve_stdio
+	try:
+		from .serve import serve_stdio
+	except ModuleNotFoundError as e:
+		if e.name != "mcp":
+			raise
+		print("camas mcp: requires feature camas[mcp]", file=sys.stderr)
+		sys.exit(2)
 
 	serve_stdio(argv)
--- a/tests/mcp/test_lazy_import.py
+++ b/tests/mcp/test_lazy_import.py
@@ -6,11 +6,13 @@ from __future__ import annotations
 import subprocess
 import sys
 from typing import TYPE_CHECKING
+from unittest.mock import patch
+
+import pytest
 
 if TYPE_CHECKING:
+	from collections.abc import Mapping, Sequence
 	from pathlib import Path
-
-	import pytest
 
 
 def test_entrypoint_mcp_branch_invokes_serve(monkeypatch: pytest.MonkeyPatch) -> None:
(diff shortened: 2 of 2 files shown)


JPHutchins:
Thanks!

If an agent generated this code, please amend the commit with the assisted by. LMK

jstar0:
Amended the commit with the assisted-by trailer and added coverage for the non-mcp missing import path. Checks are rerunning on the updated head.

PR #103, opened 2026-06-28
fix(packaging): include ty type checker in [mcp] extra

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. Add ty type checker to the [mcp] extra so camas_check works with just the mcp extra.

## Summary

`camas_check` advertises type checking but silently no-ops when installed via `camas[mcp]` because the `ty` checker isn't pulled in.

## Changes

- Add `ty>=0.0.31,<1` to the `[mcp]` extra so `uvx camas[mcp]` alone gives a fully functional MCP server

Fixes #94

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #106, opened 2026-06-28
fix(plugin): use portable uvx launcher in .mcp.json and hooks.json

### LINKED ISSUE #92: Plugin .mcp.json fails to connect: missing camas[mcp] extra and assumes global camas binary
## Summary

The Claude Code plugin's bundled `.mcp.json` fails to connect the camas MCP server out of the box. Two compounding problems:

1. **Missing `[mcp]` extra.** The `mcp` subcommand requires the optional `mcp` feature, but the plugin installs/invokes bare `camas`. Running it directly gives:
   ```
   camas mcp: requires feature camas[mcp]
   ```
2. **Assumes a `camas` binary on PATH.** The plugin config uses `command: "camas"`, but a fresh install (uvx-based, no global `camas` on PATH) has nothing to exec, so the server dies with `No such file or directory`.

Net effect: `mcp__camas__camas_gate` never registers, and the gate / `camas-fixer` agent are unavailable.

## Affected files

In the plugin source (and the regenerated cache copy):

- `agent/claude/plugin/.mcp.json`
- top-level `.mcp.json` (the marketplace one uses `uv run camas mcp --rich`, which only works inside the camas dev checkout)

Current shipped config:
```json
{
  "mcpServers": {
    "camas": {
      "type": "stdio",
      "command": "camas",
      "args": ["mcp"]
    }
  }
}
```

## Reproduction

```console
$ uvx camas mcp --rich
camas mcp: requires feature camas[mcp]

$ which camas
# (nothing)
```

## Working invocation

```console
$ uvx 'camas[mcp]' mcp --rich
Installed 30 packages in 79ms
# starts, waits on stdin
```

## Suggested fix

Ship a config that doesn't assume a global binary and pulls the extra:

```json
{
  "mcpServers": {
    "camas": {
      "type": "stdio",
      "command": "uvx",
      "args": ["camas[mcp]", "mcp", "--rich"]
    }
  }
}
```

## Notes / non-fix

For non-Python projects, the generated `tasks.py` PEP 723 path (`uv run tasks.py`) is **not** a viable launcher for the server: `run_cli(globals())` only dispatches registered task names, so `uv run tasks.py mcp` returns `no task named 'mcp'` rather than starting the server. Worth a doc note, since the `tasks.py` docstring advertises the standalone `uv run tasks.py` flow and a reader might assume it covers `mcp` too.

Also: `uvx` does not read the PEP 723 `dependencies = ["camas"]` pin in `tasks.py`, so the MCP server version would need to be pinned in `.mcp.json` directly (e.g. `"camas[mcp]>=X.Y"`) if version pinning is desired for the server.

### LINKED ISSUE #96: Plugin hooks hardcode bare `camas` — fail when camas isn't on PATH (the setup camas mcp init itself produces)
## Summary

The plugin's `hooks/hooks.json` hardcodes a bare `camas` executable:

```json
"FileChanged":    "camas mcp fix --paths ${file_path}"
"PostToolBatch":  "camas mcp gate"
```

But `camas` is frequently **not on PATH** — including in the very setup the plugin's own `camas mcp init` produces. On a uv project that has only ever run camas ephemerally, both hooks fail:

```
PostToolBatch hook error
Failed with non-blocking status code: /bin/sh: 1: camas: not found
```

## Why this is inconsistent with the rest of camas

`mcp/scaffold.py` already solves portable launching for `.mcp.json` via `launch_command()`:
- uv project (a `uv.lock` exists) → `uv run camas …`
- else camas on PATH → `camas …`
- else → refuse to write a non-portable entry

So the MCP **server** is launched portably, but the **hooks** assume a global install that `camas mcp init` never creates. A user who installs the plugin + runs `camas mcp init` (or uses a `uvx camas[all]`-style `.mcp.json`) gets working MCP tools but broken hooks — confusing, because the tools clearly prove camas is reachable.

## Repro

1. uv project, no global camas install (`uvx`/`uv run` only).
2. Install the plugin; the MCP server works (tools resolve via uvx/uv run).
3. Edit any file → `FileChanged`/`PostToolBatch` fire → `/bin/sh: camas: not found`.

## Temporary workaround (what I did)

Either install camas on PATH (`uv tool install 'camas[all]'`), or edit the cached `hooks.json` to match the `.mcp.json` launcher:

```json
"command": "uvx camas[all] mcp fix --paths ${file_path}"
"command": "uvx camas[all] mcp gate"
```

Both work, but the cache edit is clobbered on plugin update, and the PATH install undoes the deliberate ephemeral-only posture.

## Suggested real fix

`hooks.json` is static plugin JSON, so it can't run `launch_command()` at write time. A few options, roughly in order of preference:

1. **Make the hook commands zero-install and portable** — ship them as `uvx camas[all] mcp fix|gate` (mirrors what an ephemeral `.mcp.json` already does; needs nothing pre-installed).
2. **Have `camas mcp init` also write the hooks** into the project's `.claude/settings.json` using the resolved `launch_command()` (the same logic that already picks `uv run camas` vs `camas`), instead of relying on the plugin's static bare-`camas` hooks. Keeps server and hooks launched identically.
3. **Ship a tiny launcher shim** the hooks call, which does the `uv run` / PATH / `uvx` resolution at runtime.
4. At minimum, **document** that the plugin requires `camas` on PATH (e.g. `uv tool install 'camas[all]'`) and have `camas mcp init` say so.

Option 1 or 2 would make the plugin work out-of-the-box for the same environments `camas mcp init` already targets.

---
*Found while wiring the plugin into a uv-based project; happy to PR whichever direction you prefer.*

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. Fix the plugin JSON files to use portable uvx launchers instead of bare camas.

## Summary

The plugin's `.mcp.json` assumed a global `camas` binary and omitted the `[mcp]` extra. The `hooks.json` hardcoded bare `camas` which fails when camas isn't on PATH — the exact setup `camas mcp init` produces for uv projects.

## Changes

- **`.mcp.json`**: use `uvx camas[mcp] mcp --rich` — portable, pulls the MCP extra
- **`hooks.json`**: use `uvx camas[all] mcp fix|gate` — portable, all extras

Fixes #92, fixes #96

### DIFF
--- a/agent/claude/plugin/.mcp.json
+++ b/agent/claude/plugin/.mcp.json
@@ -2,10 +2,8 @@
   "mcpServers": {
     "camas": {
       "type": "stdio",
-      "command": "camas",
-      "args": [
-        "mcp"
-      ]
+      "command": "uvx",
+      "args": ["camas[mcp]", "mcp", "--rich"]
     }
   }
 }
--- a/agent/claude/plugin/hooks/hooks.json
+++ b/agent/claude/plugin/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "camas mcp fix --paths ${file_path}"
+            "command": "uvx camas[all] mcp fix --paths ${file_path}"
           }
         ]
       }
@@ -16,7 +16,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "camas mcp gate"
+            "command": "uvx camas[all] mcp gate"
           }
         ]
       }
--- a/tests/plugin/test_shipped_plugin.py
+++ b/tests/plugin/test_shipped_plugin.py
@@ -19,23 +19,26 @@ def test_plugin_manifest_is_named_camas() -> None:
 	assert json.loads((_PLUGIN / ".claude-plugin" / "plugin.json").read_text())["name"] == "camas"
 
 
-def test_bundled_mcp_server_launches_camas_mcp() -> None:
+def test_bundled_mcp_server_is_a_portable_stdio_entry() -> None:
 	server = json.loads((_PLUGIN / ".mcp.json").read_text())["mcpServers"]["camas"]
-	assert (server["command"], server["args"]) == ("camas", ["mcp"])
+	assert server["type"] == "stdio"
+	assert server["command"] in ("camas", "uvx", "uv")
+	assert "mcp" in server["args"]
 
 
-def test_gate_hook_runs_camas_mcp_gate() -> None:
+def test_gate_hook_runs_mcp_gate() -> None:
 	hooks = json.loads((_PLUGIN / "hooks" / "hooks.json").read_text())["hooks"]
 	hook = hooks["PostToolBatch"][0]["hooks"][0]
 	assert hook["type"] == "command"
-	assert "camas mcp gate" in hook["command"]
+	assert "mcp gate" in hook["command"]
 
 
-def test_filechanged_hook_runs_the_deterministic_autofix() -> None:
+def test_filechanged_hook_runs_mcp_fix() -> None:
 	hooks = json.loads((_PLUGIN / "hooks" / "hooks.json").read_text())["hooks"]
 	fc = hooks["FileChanged"][0]["hooks"][0]
 	assert fc["type"] == "command"
-	assert "camas mcp fix" in fc["command"]
+	assert "mcp fix" in fc["command"]
+	assert "${file_path}" in fc["command"]
 
 
 def test_marketplace_points_at_the_shipped_plugin() -> None:


JPHutchins:
Re-running CI — the status-modes-demo failure looks like a flaky runner (only 2 JSON files changed, no Python code touched).

JPHutchins (commented):
I'm not sure that we can do this. Do we KNOW that the environment has uvx? I suppose we could require that...

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> Posted by claude-opus-4-8 on behalf of @JPHutchins, who asked me to close this if it is superseded — it is.

Closing as **superseded by #116** (merged to `main` `bf43bbc`).

#116 landed the same fix — portable `uvx` launchers for the plugin's bundled `.mcp.json` + `hooks/hooks.json` — and **#92 and #96 are already closed** by it.

Two ways this PR is behind the merged state:

- It still wires a `gate` (PostToolBatch) hook, which #114 removed — the gate is now run by the `camas-fixer` subagent, not a hook. #116's `hooks.json` is FileChanged-only.
- It uses `camas[all]` + `--rich`; #116 uses the minimal `camas[mcp]` (which carries `ty` via #103) with no `--rich`.

It's also CONFLICTING against the now-merged files. Nothing to salvage — thanks for the fix; the problem is resolved in main.

Issue #101, opened 2026-06-28
PostToolBatch gate is synchronous — can't express the intended async background-fixer model


JPHutchins:
## Summary

The `PostToolBatch` gate is a **synchronous** command hook, but its intended design — as described by the author — is to **spawn scoped fixer subagents in parallel with the main agent, making background progress on trivial fixes without blocking**. A `PostToolBatch` command hook cannot express that: the hook runs to completion before the agent continues, and a `block` decision halts continuation outright. The async intent degrades into a stall.

## What I observed

Driving a real project, the gate (on a `needs_reasoning` residual) blocks; the residual is then delegated to the `camas:camas-fixer` subagent, which *"fixes the scoped files and re-gates that same scope in a loop"*. When the loop can't turn the residual green quickly, the whole REPL is frozen for minutes — I saw a "Crunched for 1m44s" stall, and the counter stopped. The hooks themselves are fast (~0.5–0.7s, measured); the time is spent in the blocking delegated loop downstream of a synchronous hook.

## Why it can't be fixed in the task tree

`PostToolBatch` hooks are synchronous by contract, and `decision: "block"` is defined to stop continuation. There is no "fire-and-forget, make progress in the background" affordance in a command hook, so the design intent (parallel, non-blocking fixers) is unreachable as currently wired.

## Suggested directions

1. **Detach**: the `PostToolBatch` hook spawns the fixer as a background process and immediately returns exit 0 (continue), so the main agent never blocks; fixes land on a later tick.
2. **Event-driven channel** instead of a synchronous hook for the "background fixer" role.
3. If synchronous-and-blocking is the only option, **document that explicitly**, keep the gate's scope tight (see #102), and lean on `--under` time-boxing so a blocking gate is always sub-second.

## Related

- #97 — the same gate hook emits invalid JSON on green; both are gate-hook integration problems.
- #96 — hook wiring/portability.

JPHutchins:
## Design: the gate is agent-orchestrated, not a hook

Resolving this the way the issue diagnosed it: a `PostToolBatch` **command hook is synchronous by contract** — it runs to completion before the agent continues, and `decision:"block"` halts continuation outright — so "spawn scoped fixers in parallel, make background progress without blocking" is unreachable *from a command hook*. `#107` was the right first move (remove it); this is the model that replaces it.

**Decision (chosen with @JPHutchins): pure SKILL guidance, no gate hook.**

[...]

PR #107, opened 2026-06-28
fix(plugin): disable PostToolBatch hook (synchronous API blocks agent)

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. Disable the synchronous PostToolBatch hook as a first step — cannot use a blocking API.

## Summary

The PostToolBatch command hook is synchronous by contract — it runs to completion before the agent continues. The intended design of spawning scoped fixer subagents in parallel with the main agent is unreachable with a synchronous command hook.

This is the first step: disable the PostToolBatch hook. The full fix requires the main agent to have a means of starting camas fixer subagents with scoped paths working in parallel.

## Changes

- Remove the `PostToolBatch` hook from `hooks.json`, keeping only the `FileChanged` autofix hook

Related: #101

### DIFF
--- a/agent/claude/plugin/hooks/hooks.json
+++ b/agent/claude/plugin/hooks/hooks.json
@@ -5,18 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "camas mcp fix --paths ${file_path}"
-          }
-        ]
-      }
-    ],
-    "PostToolBatch": [
-      {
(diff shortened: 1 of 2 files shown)


JPHutchins (commented):
LGTM

PR #110, opened 2026-06-28
docs: add Claude agent integration to tutorial and --init starter

### LINKED ISSUE #95: Tutorial/--init starter omit the Claude agent integration (Config(agent=Claude), AgentFormat, the two-hook gate)
## Summary

The `camas_docs` tutorial (and the `camas --init` starter it mirrors) never mentions the **Claude / agent integration**, even though the MCP server instructions tell agents to read `camas_docs` *before* authoring `tasks.py`. An agent following that guidance has no in-band documentation for the very part of the config that exists for it — so it writes a plain task tree and silently omits the gate wiring.

I exercised the plugin end-to-end (`camas_list` → `camas_run ci` → `camas_docs` → `camas_check`, all green) and only discovered the integration by reading the installed source. Everything below is present and documented *in code* — the gap is purely that the tutorial doesn't surface it.

## What's missing from the tutorial

1. **`Config(agent=Claude(fix=..., check=...))`** (`v0/config.py`)
   - `fix`: the deterministic, behavior-preserving autofix node the `FileChanged` hook runs (scoped, zero tokens). The docstring's key subtlety — *declared, not derived from `mutates=`*, because a mutating leaf may be codegen/a compiler rather than a fixer — is exactly the kind of thing an authoring agent needs told, since it would otherwise assume `mutates=True` is the fixer set.
   - `check`: the node the `PostToolBatch` gate runs; `None` defers to `default_task`/`github_task`, scoped by `--paths`, time-boxed by `--under`.

2. **The two-hook split** (`core/gate.py`): `FileChanged` mutates (`camas fix`), `PostToolBatch` only checks (`camas_gate`, read-only) → `green`/`continue` vs `needs_reasoning`/`block`. The tutorial documents `mutates=`/`--under` for the human inner loop but never connects them to the gate.

3. **`AgentFormat` / `Task(agent_format=("<flag>", "<kind>"))`** (`v0/task.py`), `OutputKind = sarif | rdjson | lsp | junit | tap | raw`. The agent-only structured-output variant `with_agent_format()` appends on gate runs but not human runs. This is genuinely non-discoverable: an agent has no reason to guess the field exists, yet it's the mechanism that gives the gate machine-readable diagnostics instead of scraped text.

## Suggested fix

Add a short tutorial section (and a commented block in the `--init` starter) showing a minimal agent-integrated config (binding names are the task names, per the existing tutorial note — `name=` only on the nested anonymous leaves):

```python
from camas import Config, Parallel, Sequential, Task
from camas.v0.config import Claude

fix   = Sequential(Task("ruff format .", name="fmt"),
                   Task("ruff check --fix .", name="lint-fix"))
check = Parallel(
    Task("ruff check .", agent_format=("--output-format sarif", "sarif")),
    Task("pytest", agent_format=("--junitxml=- ", "junit")),
)
_ = Config(default_task=check, agent=Claude(fix=fix, check=check))
```

A pointer from the existing `mutates=`/`--under` paragraph to "and here's how the gate uses these" would also help, since that's where an agent is already reading when it's thinking about hooks.

## Bonus question

`camas mcp init` / `mcp/scaffold.py` only writes the `.mcp.json` server entry — where do the `FileChanged`/`PostToolBatch` hook entries in `settings.json` come from? If the plugin install wires them, a one-liner in the tutorial confirming that (vs. "add them yourself") would close the loop.

---
*Filed after exercising the plugin in a real session; happy to PR the tutorial/starter additions if useful.*

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. Add Claude agent integration documentation to the camas_docs tutorial and --init starter template.

## Summary

The `camas_docs` tutorial and `camas --init` starter never mentioned the Claude/agent integration — `Config(agent=Claude(fix=..., check=...))`, `AgentFormat`, or the two-hook gate model.

## Changes

- **starter.py**: Added agent integration comments explaining `Claude(fix=, check=)`, `AgentFormat`, and the two-hook model
- **\_\_init\_\_.py tutorial**: Added a section on the agent integration with a worked example

Fixes #95

### DIFF
--- a/src/camas/__init__.py
+++ b/src/camas/__init__.py
@@ -329,6 +329,39 @@ https://github.com/JPHutchins/camas/blob/main/tests
 
 Effect protocol: subclass ``Effect`` (see ``examples/effect-plugin/``). Per-task
 help: ``camas <task> --help``.
+
+**Agent integration** (the camas Claude Code plugin)::
+
+    from camas import Claude, Config
+    from camas.v0.task import AgentFormat
+
+    # The deterministic, behavior-preserving autofix node the FileChanged
+    # hook runs on every edit (scoped, zero tokens). Mutates=True so
+    # --under runs it first; paths= so it scopes to the changed file.
+    autofix = Parallel(
+        Task("ruff format {paths}", mutates=True, paths="."),
+        Task("ruff check --fix {paths}", mutates=True, paths="."),
+    )
+
+    # The check node the PostToolBatch gate runs. When a leaf carries
+    # agent_format=, the gate appends those flags (machine-readable output
+    # like sarif or junit) so the agent gets structured diagnostics; a
+    # human ``camas lint`` never sees these extra flags.
+    checks = Parallel(
+        Task("ruff check .",
+             agent_format=AgentFormat("--output-format sarif", "sarif")),
+        Task("pytest",
+             agent_format=AgentFormat("--junitxml=-", "junit")),
+    )
+
+    _ = Config(default_task=checks, agent=Claude(fix=autofix))
+    # agent.check defaults to default_task; set explicitly to gate a
+    # different node: agent=Claude(fix=autofix, check=fast_checks)
+
+The plugin's ``FileChanged`` hook runs ``camas mcp fix --paths <file>`` on
+every edit. ``PostToolBatch`` runs ``camas mcp gate`` after tool batches,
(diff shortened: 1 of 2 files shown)


JPHutchins (commented):
Cannot merge for now, we're no longer using the PostBatchHook.

PR #114, opened 2026-06-30
fix(plugin): gate is agent-orchestrated, not a PostToolBatch hook (closes #97)

### LINKED ISSUE #97: PostToolBatch gate emits `decision: "continue"` on green — invalid Claude Code hook output, errors every passing run

### LINKED ISSUE #101: PostToolBatch gate is synchronous — can't express the intended async background-fixer model

JPHutchins:
## What

`#107` disabled the synchronous `PostToolBatch` gate hook — correctly: a Claude Code command hook runs to completion before the agent continues, and `decision:"block"` halts it, so it can't express the parallel-background-fixer model. But the docs (and `camas mcp init --hooks`) still **described and installed** that hook. This aligns everything with the chosen design (`#101`): **the main agent delegates the check-and-fix loop to background `camas-fixer` subagents, scoped to the changed paths, off its own context** — there is no check hook.

## Changes

- **`camas mcp init --hooks` no longer writes the `PostToolBatch` gate hook** (only the `FileChanged` autofix). This removes the install path that hit `#97` (invalid `decision:"continue"` errored every green run) and the `#101` stall. (`scaffold.py` + test.)
- **Plugin agent docs rewritten to the background-subagent model** (`SKILL.md`, `camas-fixer.md`): no check hook; after a batch of edits, delegate to `camas-fixer` with the changed paths, run it in the background, spawn one per independent scope to parallelize. Context-isolation, parallelism, and backgrounding verified against current Claude Code docs.
- **Plugin manifests** (`plugin.json`, `marketplace.json`): dropped "residual surfacing after each batch" (the removed hook) for the delegated fixer subagent.
- **De-hooked the gate descriptions/docstrings** that framed it as PostToolBatch-driven (`serve.py`, `wire.py`, `cli.py`, `core/gate.py`, `core/scope.py`). The gate's *behavior* is unchanged; only the framing.

## Validation

`camas run check` (format, lint, mypy, pyright, ty, zuban, pyrefly, tests + doctests) and `camas run coverage` (100%) both green.

Closes #97. Closes #101.

> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-4-8 on behalf of @JPHutchins. June asked me to survey the repo, resolve #97, implement the #101 fixer model she chose (pure SKILL guidance, no gate hook), and audit the plugin/agent docs for SSOT drift — which surfaced the scaffolder and doc references this PR corrects.

https://claude.ai/code/session_01C4BSx9623GnyfigqjnEjFX

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,7 +7,7 @@
     {
       "name": "camas",
       "source": "./agent/claude/plugin",
-      "description": "The camas gate \u2014 deterministic autofix on every change, plus residual-only surfacing after each batch."
+      "description": "The camas gate \u2014 deterministic autofix on every change, plus a cheap fixer subagent that drives changed files to green off the main agent's context."
     }
   ]
 }
--- a/agent/claude/plugin/.claude-plugin/plugin.json
+++ b/agent/claude/plugin/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "camas",
-  "description": "The camas gate: applies the deterministic fixes for free after each change, and surfaces only the residual that needs reasoning.",
+  "description": "The camas gate: applies deterministic fixes for free on every change, and the check-and-fix loop is delegated to a cheap fixer subagent, so only the residual that needs reasoning reaches the main agent.",
   "version": "0.1.0"
 }
--- a/agent/claude/plugin/agents/camas-fixer.md
+++ b/agent/claude/plugin/agents/camas-fixer.md
@@ -1,20 +1,20 @@
 ---
 name: camas-fixer
-description: Resolves a camas gate `needs_reasoning` residual on a cheap model, off the main agent's context — fixes the scoped files and re-gates that same scope in a loop within a bounded turn budget, then hands any unresolved residual back. Invoke it when the gate surfaces a residual, passing the gate's rerun scope (its changed paths).
+description: Drives a scope of changed files to a green camas gate on a cheap model, off the main agent's context — gates the scope, fixes what fails, and re-gates in a bounded loop, then hands back any residual it cannot settle. Delegate to it after a batch of edits with the changed paths as its scope; spawn one per independent scope to run them in parallel, and run it in the background so it never blocks your reasoning.
 model: haiku
 maxTurns: 5
 tools: Read, Edit, Bash, mcp__camas__camas_gate
 ---
 
-You resolve camas gate residuals cheaply, so the main agent spends no reasoning on what a
-fix-and-recheck loop can settle. You are given the failing diagnostics and the scope that
-produced them — the gate's `rerun` handle: the changed paths (and task) it ran over. Loop:
+You drive a scope of changed files to a green camas gate cheaply, so the main agent spends no
+reasoning on what a fix-and-recheck loop can settle. You are given the changed paths to work
+over (and, when the main agent already ran the gate, its failing diagnostics). Loop:
 
-1. Read the diagnostics, edit the code to fix the underlying cause, and re-gate the SAME scope:
-   run `camas mcp gate --paths <the changed paths you were given>` (do not widen the scope —
-   re-gate exactly what was handed to you).
+1. Gate the scope: run `camas mcp gate --paths <the changed paths you were given>` to see what
+   fails — do not widen the scope, gate exactly what was handed to you.
 2. If it comes back green (exit 0), you are done — say so.
-3. If it still fails, refine and re-gate. Repeat until green or you run out of turns.
+3. If it still fails, read the diagnostics, edit the code to fix the underlying cause, and
+   re-gate the same scope. Repeat until green or you run out of turns.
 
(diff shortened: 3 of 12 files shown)


JPHutchins (commented):
LGTM

PR #149, opened 2026-07-02
feat(scope): paths on Group + correct {paths}-less run rule; #141 docs papercuts

### LINKED ISSUE #141: cleanup: docs & minor housekeeping (--under help, starter example, gate dry-run, test comments)
> [!WARNING]
> **LLM Disclosure**
>

**Part of #128 (epic).** A bundle of low-severity papercuts (each a small, independent fix). Confidence: **confirmed**.

- [ ] **E1 · `--under` help is stale.** `src/camas/main/parser.py:315` still says "untimed leaves are skipped," but the in-window budget change now *runs* untimed leaves (to measure them). Also document the behavior shift: `--under` on a **cold** timing cache now runs the full tree — "bounded" degrades to "full first run" — which surprises a caller relying on `--under` as a hard ceiling. Update the help string and the relevant README/docstring note.

- [ ] **E2 · Starter template's commented example uses a non-existent API.** `src/camas/main/starter.py:54` (inside a comment) shows `Parallel(Task(…), Task(…), paths=".")`, but `paths` is a `Task`-only field — `Group` has no `paths`. A user adapting that example hits `TypeError`; and the two inner `Task`s carry `{paths}` in `cmd` with no `paths=` declaration, so they wouldn't scope either. Fix the commented example to a valid form (per-`Task` `paths=`). (Live code on line 55 is a valid `Task`, so `camas --init` output itself is fine.)

- [ ] **E3 · `run_gate_cli --dry-run` ignores `--under`.** `src/camas/mcp/serve.py` — the dry-run branch scopes and prints the plan but skips `plan_under`, so `camas mcp gate --dry-run --under 5s` previews leaves the real run would exclude (and in the wrong scheduling order). Apply the budget in the preview so dry-run matches the real gate.

- [ ] **E4 · Inline explanatory comments in new tests.** `tests/core/test_gate.py:43,59` and `tests/main/test_dispatch.py:299` add why-prose the assertions already encode — against the repo convention to avoid inline comments. Remove or fold into test names. Non-blocking.

## Acceptance

- Each checkbox resolved; `--under` help matches behavior; the starter's commented example is valid; gate dry-run reflects `--under`.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-4-8[1m] on behalf of @JPHutchins. It began as the #141 documentation papercuts; on her feedback the E2 item turned out to be a missing feature (`paths` on `Group`), which is implemented here, along with a `Group` pattern-matching cleanup she asked for.

Closes #141.

## `paths` on `Group` (was E2)

E2 was filed as "the starter's commented example uses a non-existent API" (`Parallel(…, paths=".")`). It's not non-existent — `paths=` was always meant to be valid on `Sequential`/`Parallel`, scoping the whole node. So this **implements** it rather than editing the comment:

- `Group` gains a `paths` field. A group's `paths` is **baked into descendant leaves by `expand_matrix`** (own wins, else inherit) — exactly how `env`/`cwd` already propagate — so a `{paths}` child without its own `paths` inherits the group's default target. This survives `--under` budget regrouping (which flattens the tree) because it's resolved during expansion, before scoping.
- The run/skip rule is corrected to match the design: **a command with no `{paths}` can't be narrowed, so it always runs** (its `paths`, own or inherited, is a no-op); only a `{paths}` command is pruned, and only when its scope covers none of the changed files. This overturns the old `mypy`/`cargo` "run-whole-or-skip" selector — those must run (err on correctness: the tool might touch the edited files). The `test_prefix_without_placeholder` test and the `cargo` doctest that pinned the old behavior are rewritten.
- The starter's `Parallel(Task("ruff format {paths}", …), Task("ruff check --fix {paths}", …), paths=".")` example is restored (now valid, and a showcase of the feature).
- `paths` carries through every group reconstruction site (matrix expand/specialize/override, scope, gate, tasks naming/promote, expression eval/serialize) so it round-trips.

## `Group` pattern-matching cleanup

The duplicated `case Sequential(...)` / `case Parallel(...)` reconstruction cases collapse to a single `case Group() as g: type(g)(...)` — cast-free, because matching `Group()` against a `TaskNode` subject narrows `g` to `Sequential | Parallel`, so `type(g)(…)` is a valid `TaskNode`. Applied in `scope`, `gate`, `override`, `promote`, `resolve_refs`, and `to_expression` (via `type(g).__name__`). The cases that behave differently per subclass (`execution`, `expand_matrix`'s helper dispatch) stay split.

## Retained papercuts

- **E1** — corrected the stale "`--under` skips untimed leaves" claim in `parser.py`, `wire.py`, and the `camas_docs`-served module docstring.
- **E3** — `camas mcp gate --dry-run` now applies `--under` (mirrors `run_gate`), previewing exactly what the real gate would run.
- **E4** — removed inline why-comments in the touched tests.

## Verification

`camas all` — lint, format, mypy, pyright, ty, zuban, pyrefly, **coverage 100%** — all green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,15 +216,17 @@ Durations are `1s`, `500ms`, `2m`, `1h`, or a bare number of seconds. Only leave
 
 ## Path scoping (`--paths`)
 
-`camas <task> --paths <path>…` scopes a run to changed paths instead of the whole tree. A leaf opts in by writing the `{paths}` placeholder in its command and declaring its scope with `paths=` — a directory prefix, or a `(changed) -> paths` callable:
+`camas <task> --paths <path>…` scopes a run to changed paths instead of the whole tree. A command opts in by writing the `{paths}` placeholder and declaring its scope with `paths=` — a directory prefix, or a `(changed) -> paths` callable. A `Sequential`/`Parallel` may carry `paths=` too: it's the default scope for descendant `{paths}` commands that set none (the same way `env`/`cwd` propagate into a group's leaves):
 
 ```python
 py = Task("ruff format {paths}", mutates=True, paths="src")
 web = Task("prettier --write {paths}", mutates=True, paths="web")
-_ = Config(agent=Claude(fix=Sequential(py, web)))
+# the group's paths="." is the default for both children (neither sets its own):
+autofix = Parallel(Task("ruff format {paths}"), Task("ruff check --fix {paths}"), paths=".")
+_ = Config(agent=Claude(fix=Sequential(py, web, autofix)))
 ```
 
-`--paths` works on any task — `camas check --paths src/a.py`. Without it, every `{paths}` resolves to its full-run default (`ruff format src`); with it, each leaf runs only over the changed files it covers, and a leaf that covers none is dropped (`--paths` is repeatable, or comma-separated).
+`--paths` works on any task — `camas check --paths src/a.py`. Without it, every `{paths}` resolves to its full-run default (`ruff format src`); with it, each `{paths}` command runs only over the changed files it covers, and one covering none is dropped. A command **without** `{paths}` can't be narrowed, so its `paths=` is a no-op and it always runs — camas errs on correctness (a tool it can't narrow might be affected by the edit). `--paths` is repeatable, or comma-separated.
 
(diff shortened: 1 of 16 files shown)


JPHutchins:
LGTM

Issue #128, opened 2026-07-02
cleanup: post-gate-epic hardening — 15 items from a 7-day review (2026-06-25 → 07-01)


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-4-8[1m] on behalf of @JPHutchins. She asked for a `/code-review` over the last 7 days of `main` (2026-06-25 → 07-01) plus the issues those commits resolved, checking that the approach stayed sound and goal-aligned, with the output filed as a cleanup issue.

## Scope

Review of the 15 PRs merged to `main` in the window (`020908e` → `3499194`): the SA-delegation gate epic landing (#89), the agent-orchestration pivot (#107/#114), plugin packaging (#116/#126), the MCP surface (#105/#108/#111/#118), the `Ctrf` effect (#120), the agentic CI review (#121), the Nix extras derivation (#122), and docs (#117/#119). Findings below were produced by fan-out finders and then verified against the current code, the Claude Code hook docs, and the `#126` commit history.

## Tracked child issues — ✅ ALL RESOLVED (see closing comment)

**Correctness / behavior**
- [x] #129 — A1 · FileChanged `${file_path}` autofix — done (#142); `PostToolBatch` on stdin
- [x] #130 — A2 · `HookCommand(extra="forbid")` aborts `camas mcp init --hooks` — done (#143); `extra="allow"`
- [x] #144 — A2-followup · `write_hooks` rewrites settings.json non-faithfully — done (#150); mutate parsed `raw` via typed helpers + `json.dumps`
- [x] #131 — A3 + A5 · `to_expression` round-trip fidelity (group fields + tuple commands) — done (#145)
- [x] #132 — A4 · path-scope input normalization (empty entry → whole tree, comma-split) — done (#148)

**Plugin / packaging / SSOT**
- [x] #133 — B1 · `camas-fixer` portability + prefer the `camas_gate` MCP tool (→ #127) — done (#150)
- [x] #134 — B2 · derive mypyc effect-exclusion from pyproject — done (#152); guard test in `tests/test_packaging.py`
- [x] #135 — B3 · guard Nix `resolveExtra` name → nixpkgs-attr mapping — done (#152)

**Agentic CI review (#121)**
- [x] #136 — C1 + C3 · keep secrets/token out of untrusted Phase 2 _(high)_ — closed **not-planned** (risk accepted); zero-cost hardening (`persist-credentials:false` + output scrub) landed in #151
[...]

PR #152, opened 2026-07-02
test(packaging): guard mypyc effect-exclusion drift; harden nix resolveExtra (#134) (#135)

### LINKED ISSUE #134: Derive mypyc effect-exclusion set from pyproject (setup.py SSOT drift)
> [!WARNING]
> **LLM Disclosure**
>

**Part of #128 (epic).** Severity: **medium** · Confidence: **confirmed**. Same drift class as #122 (Nix extras) and #126 (plugin version).

## Problem

`setup.py:26` hardcodes the set of effect modules that must stay interpreted under mypyc:

```python
_effect_excluded = {"github_checks.py", "ctrf.py"}
```

This is a hand-maintained mirror of "effects with an optional dependency" (httpx → `github_checks`, msgspec → `ctrf`): their optional deps aren't in the isolated build env, so mypyc can't compile them. Every new optional-dep effect must be added here or the compiled (`CAMAS_USE_MYPYC=1`) build breaks — the same "value duplicated off pyproject" pattern the Nix-extras (#122) and plugin-version (#126) fixes just eliminated.

## Action

Pick one:
- **Derive** the exclusion from the pyproject `[project.optional-dependencies]` (the effect modules whose dep lands behind an extra), the way #122 derives the Nix extras — one source of truth.
- **Or** add a test asserting every effect module backed by an `[extra]` appears in `_effect_excluded`, so the drift fails CI loudly instead of at build time.

## Acceptance

- Adding a new optional-dep effect does not require editing `setup.py` by hand (derivation), or fails a fast test if it would break the compiled build (guard).

### LINKED ISSUE #135: Guard Nix `resolveExtra` package-name to nixpkgs-attr mapping
> [!WARNING]
> **LLM Disclosure**
>

**Part of #128 (epic).** Severity: **low** (maintainability) · Confidence: **confirmed** (not triggered by today's extras).

## Problem

`nix/package.nix` `resolveExtra` — #122 correctly derives the extra *names* from pyproject, but each dependency spec is mapped to `python3Packages.${name}` by naive identity. This has three unguarded edges for future pyproject edits:

1. An extra whose PyPI distribution name ≠ its nixpkgs attr name throws an opaque eval error.
2. An extra listing a package not in nixpkgs (e.g. `ty`, `mcp`) throws on the `python3Packages.${name}` lookup.
3. A mutually self-referential extra pair (`a = ["camas[b]"]`, `b = ["camas[a]"]`) recurses infinitely.
4. `builtins.head (builtins.match …)` on a spec that doesn't start with `[A-Za-z0-9._-]` (a VCS/URL/`@`-marker requirement) returns `null`, and `builtins.head null` throws.

None fire for the current extras, but this file is now the SSOT for extras, so a future edit could fail cryptically at eval time.

## Action

- Guard the mapping: an explicit PyPI-name → nixpkgs-attr table (or `lib.attrByPath` with a clear assertion), and a null-match guard on `parseSpec`, so a future extra fails loudly with a readable message rather than an opaque eval trace or infinite recursion.

## Acceptance

- Adding an extra whose package isn't resolvable produces a clear eval-time error naming the extra/package.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-4-8[1m] on behalf of @JPHutchins, bundling the packaging-SSOT items of the #128 cleanup epic. It was implemented and validated (`camas all`) in an isolated worktree.

Closes #134, closes #135. Two packaging-SSOT drift guards.

## #134 — mypyc effect-exclusion drift (guard test)

Per maintainer decision this is a **guard test**, not a derivation in `setup.py` (which is excluded from all five typecheckers and coverage — the derivation would ship untested). New `tests/test_packaging.py` derives the expected exclusion set from the **real effect-module imports** + pyproject extras and asserts it equals `setup.py`'s `_effect_excluded` literal, so drift fails the fast `check` job before the slow wheels build.

Detection scans actual imports rather than the brittle extra-name==basename coincidence: optional dist names = union of every `[project.optional-dependencies]` spec (minus `camas[...]` self-refs) minus the core deps → `{httpx, msgspec, ty, mcp}`; a public effect module is excluded iff its own or a `from . import _sibling` private sibling's import roots intersect that set — so `github_checks.py` matches via its own `httpx`, `ctrf.py` via `_ctrf_model`→`msgspec`. Helpers stay comprehension/`match`-based with the `tomllib`/`tomli` shim, holding 100% branch coverage. `setup.py` is unchanged. (Verified the guard fires by temporarily dropping `ctrf.py` from the literal.)

## #135 — Nix `resolveExtra` guards

Extracted `parseSpec` + `resolveExtra` from `nix/package.nix` into a pure `nix/resolve-extras.nix` (a `mkResolveExtra` factory) with four readable-throw guards: an explicit `pypiToNixAttr` override table + a has-attribute assertion (PyPI≠nixpkgs-attr / package absent), cycle detection via a threaded `seen` list, and a null-match guard in `parseSpec`. Behavior is identical for today's extras. A new flake `checks.extras-resolver` asserts the real extras resolve and that synthetic cyclic / missing-package / unparseable-spec inputs each fail eval (`builtins.tryEval`); `resolve-extras.nix` was added to the `nixfmt --check` command.

## Validation

`camas all` green (100% coverage, all typecheckers). The Nix `checks.extras-resolver` runs under the existing `nix flake check` CI job.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -89,9 +89,60 @@
                 nativeBuildInputs = [ pkgs.nixfmt ];
               }
               ''
-                nixfmt --check ${./flake.nix} ${./nix/package.nix}
+                nixfmt --check ${./flake.nix} ${./nix/package.nix} ${./nix/resolve-extras.nix}
                 touch $out
               '';
+
+          extras-resolver =
+            let
+              resolver = import ./nix/resolve-extras.nix {
+                inherit lib;
+                pname = "camas";
+              };
+              realExtras = (lib.importTOML ./pyproject.toml).project.optional-dependencies;
+              resolveNames =
+                pyprojectExtras: python3Packages: extra:
+                map (drv: drv.name) (resolver.mkResolveExtra { inherit python3Packages pyprojectExtras; } extra);
+              forced = value: builtins.tryEval (builtins.deepSeq value value);
+              fakePackages = {
+                httpx = {
+                  name = "httpx";
+                };
+                msgspec = {
+                  name = "msgspec";
+                };
+              };
+              realResolved = builtins.mapAttrs (
+                extra: _: resolveNames realExtras pkgs.python3Packages extra
+              ) realExtras;
+              realResolves = (forced realResolved).success;
+              cyclicFails =
+                !(forced (
+                  resolveNames {
+                    a = [ "camas[b]" ];
+                    b = [ "camas[a]" ];
+                  } fakePackages "a"
+                )).success;
+              missingFails =
+                !(forced (
+                  resolveNames {
+                    x = [ "definitely-absent-pkg" ];
+                  } fakePackages "x"
+                )).success;
+              unparseableFails =
+                !(forced (
+                  resolveNames {
+                    y = [ "@vcs+https://example/x" ];
+                  } fakePackages "y"
+                )).success;
+            in
+            assert realResolves;
+            assert cyclicFails;
+            assert missingFails;
+            assert unparseableFails;
+            pkgs.runCommand "extras-resolver-check" { } ''
+              touch $out
+            '';
         }
         // lib.mapAttrs' (
           name: pkg: lib.nameValuePair ("package" + lib.optionalString (name != "default") "-${name}") pkg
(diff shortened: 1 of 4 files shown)


JPHutchins (commented):
OK

JPHutchins:
LGTM

PR #153, opened 2026-07-02
fix(effect): no effect crashes on a zero-leaf run; Ctrf $schema at root (#140)

### LINKED ISSUE #140: Ctrf: crash on zero-leaf run + `$schema` emitted under extra instead of root
> [!WARNING]
> **LLM Disclosure**
>

**Part of #128 (epic).** Two `Ctrf` (#120) fixes. Confidence: **confirmed**.

## D1 · Zero-leaf run crashes in teardown — severity: **medium**

`src/camas/effect/ctrf.py:101` reads `ctxs[0]` unconditionally. A run with zero leaves — an empty matrix axis (`matrix={"X": ()}` expands to 0 leaves) or `Parallel()` — makes `run(…, effects=(Ctrf(),))` raise `IndexError` in teardown, aborting the whole run with a `BaseExceptionGroup("teardown errors", …)`. For an effect whose entire purpose is to be a CI artifact, an empty axis is a realistic misconfiguration that should produce an empty-but-valid report, not a crash.

(`Summary.teardown`, `src/camas/effect/summary.py:116`, shares the `ctxs[0]` pattern — consistent with existing code, but `Ctrf` is what makes the empty case reachable in a real CI run.)

**Action:** guard the empty-`ctxs` case in `Ctrf.teardown` and emit a valid report with zero tests. Consider fixing `Summary.teardown` similarly.

## D2 · `$schema` emitted under `extra` instead of the CTRF root property — severity: **low**

`src/camas/effect/_ctrf_model.py` nests the CTRF schema URL as `extra.$schema`. CTRF defines `$schema` as a **root-level** report property; nesting it under `extra` still validates (extra is free-form) but schema-aware tooling (VS Code, ctrf validators) keys off the root and won't associate the report with its schema, so it isn't self-describing.

**Action:** emit `$schema` at the report root; update the test that currently asserts `report["extra"]["$schema"]`.

## Acceptance

- `camas <task-with-empty-axis> --effects='(Ctrf(),)'` writes a valid empty report instead of crashing.
- The emitted report carries a root-level `$schema` recognized by CTRF validators.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-4-8[1m] on behalf of @JPHutchins, implementing the zero-leaf teardown fix of the #128 cleanup epic. It was implemented and validated (`camas all`) in an isolated worktree.

Closes #140.

## Root cause

On a zero-leaf run (an empty matrix axis `matrix={"X": ()}` → 0 leaves, or `Parallel()`) the engine still calls `setup()` and `teardown()` for every active effect, but the ctx grid has one row per leaf — so teardown receives an **empty tuple**. Every affected teardown indexed `ctxs[0]` and raised `IndexError`, wrapped as `BaseExceptionGroup("teardown errors")`, aborting the whole run.

## Fix — all four affected effects (maintainer decision: not Ctrf-only)

`camas <empty-matrix-task>` crashed *today* via the default interactive renderer, independent of Ctrf, so per maintainer decision this fixes the whole class:

- **Ctrf** — reconstructs the empty case in teardown with no stashed state: `(states, start_ms)` = `(ctxs[0].state.states, ctxs[0].start_ms)` when non-empty, else `((), now_ms())`; emits a valid empty report (`summary.tests == 0`, `results.tests == []`).
- **Summary / Termtree / Timings** — `setup()` stashes its context on `self`; teardown uses `ctxs[0] if ctxs else self._ctx` and renders an empty result from the stashed context. Termtree cancels its tick task via the stashed context so no asyncio task leaks.

## Ctrf D2 — `$schema` at the report root

Moved the schema URL to the CTRF report root via `schema_url: str = msgspec.field(name="$schema")` as the first `Report` field; deleted the unused `ReportExtra` struct and `Report.extra`. `build()` sets `schema_url=SCHEMA_URL`. Per-test `LeafExtra`/`Test.extra` untouched. The emitted key moves from `report["extra"]["$schema"]` to `report["$schema"]`.

## Tests

Per-effect zero-leaf tests (drive over `Parallel()`) asserting valid empty results and no raise; the Ctrf test asserts root `$schema`, `reportFormat == "CTRF"`, `summary.tests == 0`, `results.tests == []`. An acceptance test in `test_execution.py` runs an empty-matrix-axis task under all four effects and asserts a clean return (no `BaseExceptionGroup`). Branch coverage stays 100%.

### DIFF
--- a/src/camas/core/execution.py
+++ b/src/camas/core/execution.py
@@ -268,9 +268,8 @@ async def run(
 		e for e, r in zip(effects, setup_results, strict=True) if not isinstance(r, BaseException)
 	)
 	setup_errors: Final = tuple(r for r in setup_results if isinstance(r, BaseException))
-	ctx_grid: Final[list[list[Any]]] = [
-		[r for r in setup_results if not isinstance(r, BaseException)] for _ in leaf_infos
-	]
+	active_ctxs: Final[list[Any]] = [r for r in setup_results if not isinstance(r, BaseException)]
+	ctx_grid: Final[list[list[Any]]] = [list(active_ctxs) for _ in leaf_infos]
 	states: Final[list[LeafState]] = [Waiting(info.task) for info in leaf_infos]
 
 	async def dispatch(leaf_idx: int, event: TaskEvent) -> None:
@@ -316,7 +315,7 @@ async def run(
 			r
 			for r in await asyncio.gather(
 				*(
-					effect.teardown(tuple(row[effect_idx] for row in ctx_grid))
+					effect.teardown(tuple(row[effect_idx] for row in (ctx_grid or [active_ctxs])))
 					for effect_idx, effect in enumerate(active_effects)
 				),
 				return_exceptions=True,
--- a/src/camas/effect/_ctrf_model.py
+++ b/src/camas/effect/_ctrf_model.py
@@ -77,12 +77,6 @@ class Test(msgspec.Struct, omit_defaults=True, rename="camel"):
 	extra: LeafExtra | None = None
 
 
-class ReportExtra(msgspec.Struct, rename={"schema_url": "$schema"}):
-	"""The top-level ``extra``, carrying the schema reference."""
-
-	schema_url: str
-
-
 class Results(msgspec.Struct):
 	"""The CTRF ``results`` object."""
 
@@ -94,11 +88,11 @@ class Results(msgspec.Struct):
 class Report(msgspec.Struct, omit_defaults=True, rename="camel"):
 	"""The top-level CTRF report document."""
 
+	schema_url: str = msgspec.field(name="$schema")
 	report_format: Literal["CTRF"]
 	spec_version: str
 	results: Results
 	generated_by: str | None = None
-	extra: ReportExtra | None = None
 
 
 def cmd_str(task: Task) -> str:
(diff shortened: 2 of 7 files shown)


JPHutchins (commented):
I'm concerned but this self._ctx assignment, I thought that it was passed by value. Why is it being assigned and then also returned? Is it None or not? Seems like not.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was posted by claude-opus-4-8[1m] on behalf of @JPHutchins, who flagged that the zero-leaf teardown fix stashed setup context on mutable `self._ctx`.

Reworked to fix this at the source rather than per-effect. You were right that the context is already passed by value — the only gap was that `execution.py` builds the teardown ctx grid with one row per leaf, so a zero-leaf run handed `teardown` an empty tuple. It now falls back to the setup-context row when there are no leaves (`ctx_grid or [active_ctxs]`), so every `teardown` gets `ctxs[0]` as before.

All the `self._ctx` stashing in Summary/Termtree/Timings and the Ctrf ternary are reverted — those four effects are byte-identical to `main` again. Only `execution.py` (the 2-line engine fix) and `_ctrf_model.py` (the unrelated `$schema`-at-root, D2) change now. `camas all` green, 100% coverage. Force-pushed as `c0ea22f`.

JPHutchins:
LGTM

PR #174, opened 2026-07-05
feat: field-report batch — when= scoping, MCP + nix hardening (#167, #169–#173)

### LINKED ISSUE #167: Friction report: matrix args, large failure output, run-log accumulation
> [!WARNING]
> **LLM Disclosure**
>
> This issue body was edited by deepseek-v4-pro on behalf of JPHutchins. Consolidated friction notes from a multi-hour camas MCP session on jphfmt (PR https://github.com/JPHutchins/jphfmt/pull/7), restructuring the seeded notes into a prioritized report with corrected framing and findings from the delegated subagents.

---

# camas friction report — Tier 3 test-hardening session (jphfmt)

Friction, gaps, and rough edges encountered driving camas MCP across a multi-hour session on [JPHutchins/jphfmt](https://github.com/JPHutchins/jphfmt) (PR [#7](https://github.com/JPHutchins/jphfmt/pull/7)): implementing a Tier 3 fuzz-equality feature with ~3 delegated subagents and dozens of `camas_run`/`camas_gate` calls. Notes are honest about what was camas's fault vs. mine.

### 1. No ergonomic way to run a single test by name through a matrix `Parallel`

The `test` task is a `Parallel` over the MSRV matrix (`cargo +stable test` + `cargo +1.88 test`). `camas_run(task="test", args=["--","whitespace_mutants_..."])` is rejected with `pass-through args (--) only apply to Task, got Parallel` — `args` only forward to a single leaf `Task`, composites reject them.

**Verdict (corrected):** this was largely my mistake. I should have authored a bare leaf task for the scoped concern (or pinned a matrix axis) instead of inlining a matrix `Parallel` and then complaining I couldn't forward args into it — self-inflicted. There's an open issue for forwarding args through composites / matrix-axis passthrough, which would also help. The workaround (a dedicated leaf task per scoped concern) is on me to author.

> **JP's note:** the agent should always have a bare task for each scoped concern. It must have inlined one and then claimed it couldn't use it because it shot itself in the foot. There's an open issue for this; matrix might work for it.

For larger suites where the full matrix isn't sub-second, this becomes a real workflow gap — either forward `args` through a `Parallel`/`Sequential` that resolves to a single leaf, or document that single-test-by-name runs require a dedicated leaf task.

### 2. Large failure output is unwieldy in the inline JSON

When an assertion includes a large string (e.g. `golden_is_a_fixpoint` diffs the entire `golden.c` file — `left` and `right` are each the full file, escaped), `camas_run(task="test")` returns a ~78KB JSON result with `truncated: true`. The data is correct and complete (full `left`/`right`, persisted to a `.log` file with a path), but it's hard to scan visually — I had to `Read` the persisted file to see the actual ~5-line diff (`int(*cb)` vs `int (*cb)`).

**Verdict:** camas faithfully relaying cargo's output, not a bug. But a diff-aware truncation mode (keep the assertion message, drop the multi-KB escaped string body, or render a unified diff) would materially improve the inner loop. The failure message is the valuable part; the two full-file bodies are noise.

### 3. `.camas/runs/` accumulation across many short runs

Each `camas_run` creates a new numbered directory under `.camas/runs/<task>/` (one session did ~20 per-shape verification runs → 20 dirs, each with a `.log` per leaf). Each leaf's full output is written to its `.log` file AND inlined (truncated) in the JSON result — two copies of ~the same output.

**Verdict:** minor. `.camas/` is gitignored and is the timing cache too, so it has to exist. But a `camas prune` command or auto-rotation of old runs would help over a long session.

## What worked well (worth saying)

- `camas_run(task="test"|"clippy"|"rust_fmt"|"check")` was intuitive and matched `camas_list` task names; sub-second returns after warm compile; clear `returncode`/`elapsed`/`passed`/`failed` JSON per leaf; `log` path available when needed.
- The matrix (stable + 1.88) ran both toolchains in parallel; results consistent across both.
- The `check` gate (11 leaves: rust_fmt_check, clippy, test×2, doc, ts_fmt_check, eslint, ts_typecheck, ts_build, knip, typos) ran green in ~1.8s and the per-leaf JSON made it trivial to confirm every gate.
- Across ~20+ sequential `camas_run` calls by a backgrounded subagent: no state issues, no stale caches, no weird interactions. Reliable.
- The deterministic failure output (inline `left`/`right` with shape name + mutant text + index) was exactly what was needed for the per-shape verification loop.

## Non-friction findings (for context)

- The whole point of the session was that camas-driven fuzz testing surfaced real formatter bugs (jphfmt normalization gaps: trailing ws before `;`, spacing around `=`, `ident (` tightening, cast open-paren) that the existing random-byte proptest had missed. The camas `test` task was the engine for that — every fix was verified by re-running the fuzz through `camas_run`. The friction above is minor relative to that value.

### LINKED ISSUE #169: with-mcp flake package: mkShell leaks PYTHONPATH and silently crashes `camas mcp` (agent falls back to CLI)
**Context.** Wiring camas 0.1.21 into a C23/embedded repo built under nix. The devShell adds `camas.packages.${system}.with-mcp` to a plain `pkgs.mkShell { packages = [ … ]; }`, and the generated `.mcp.json` launches the server with `uvx 'camas[mcp]' mcp`.

## Observed

Inside that shell, `mkShell` sets `PYTHONPATH` to camas's entire python3.13 closure (`camas` + `mcp` + `pydantic` + `pydantic-core` + …). Both `camas mcp …` and `uvx 'camas[mcp]' mcp …` then die on import:

```
ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'
```

The leaked python3.13 `pydantic_core` shadows the isolated one uv/uvx set up, and its compiled extension doesn't match the running interpreter. The plain task runner (`camas <task>`) is unaffected — only the mcp/pydantic path breaks.

## Why it's nasty: it fails *silently*

When Claude Code launches the server at session start and it crashes this way, the agent sees **no `mcp__camas__*` tools and no error** — indistinguishable from "no MCP configured." It just quietly uses the `camas` CLI for the entire session (exactly what happened to me), and the `camas-fixer` subagent is dead too since it shares the server. There's no in-session recovery — the server is spawned once at launch.

## Workaround (confirmed fix)

`shellHook = "unset PYTHONPATH";` in the consumer devShell — the same thing camas's *own* flake does. After that, `uvx 'camas[mcp]==0.1.21' mcp gate` returns `residual_class: green`.

## Suggestions

1. **Robustness** — make the `with-mcp` app immune to an ambient `PYTHONPATH`: wrap the entry point with `--unset PYTHONPATH` (or run the interpreter with `-I`/`-S`) so `mkShell` leakage can't shadow its deps. A flake consumer shouldn't have to know this.
2. **Docs** — a consumer note: "`unset PYTHONPATH` in your devShell, or the MCP server won't start."
3. **Diagnosability** — preflight the critical imports (`mcp`, `pydantic_core`) and emit one actionable stderr line on failure, and/or add a cheap `camas mcp doctor` health subcommand, so a dead server isn't invisible.

## Related

- #127 — "agent used CLI instead of MCP despite both being active." This is one concrete root cause: the server had crashed but *looked* configured (plugin files all present), so the agent had no tools and fell back to the CLI.
- #58 — MCP server lifecycle / staleness.

> [!WARNING]
> **LLM Disclosure**
>
> This issue was authored by `claude-opus-4-8` on behalf of @JPHutchins, who asked me to wire camas 0.1.21 (with its MCP + Claude Code plugin) into a nix-built C23 repo and file the friction I hit. This one cost a whole session of CLI-only fallback before I traced it.

### LINKED ISSUE #170: camas mcp init writes an unpinned `uvx 'camas[mcp]'` launcher — version can drift
## Observed

`camas mcp init --claude` writes an **unpinned** launcher in two places:

- `.mcp.json` → `uvx camas[mcp] mcp`
- the PostToolBatch hook (`.claude/settings.json`) → `uvx 'camas[mcp]' mcp fix`

Both float to whatever `uvx` resolves for `camas` at run time — independent of the version the project was set up with, and independent of a version a flake pins. So the MCP server / autofix hook can silently drift to a newer camas than the rest of the project expects.

The README describes a "pinned launcher that resolves camas from the tasks.py PEP 723 block," but that isn't what lands here.

## Suggestion

- Pin at `init` time — write the resolved version, e.g. `uvx 'camas[mcp]==0.1.21' mcp` (this is what I did by hand).
- And/or a `--launcher` choice: `uvx` (pinned) | a PATH-provided `camas` (e.g. from a flake devShell) | explicit path — so a nix/flake-provided camas can back the server instead of a second PyPI fetch.

## Related

- #58 — MCP server serving stale code; a pinned launcher is adjacent to that class of "which camas is actually running?" problem.

> [!WARNING]
> **LLM Disclosure**
>
> This issue was authored by `claude-opus-4-8` on behalf of @JPHutchins, who asked me to set up camas 0.1.21 specifically (via a flake) and file friction found along the way; the unpinned launcher meant the MCP could run a different version than the pinned flake.

### LINKED ISSUE #171: Flake with-mcp reports 0.0.0+<shortRev> for a tagged build
## Observed

Consuming the flake at a release tag still reports a `0.0.0` version:

```console
$ nix run 'github:JPHutchins/camas/0.1.21#with-mcp' -- --version
camas 0.0.0+09f0fca
```

`flake.nix` computes:

```nix
version = "0.0.0+${self.shortRev or self.dirtyShortRev or "unknown"}";
```

and feeds it to `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_CAMAS`, so even a build from the `0.1.21` tag loses the real version — it becomes `0.0.0+<shortRev>`.

## Suggestion

Derive the version from the tag when the source is at one — e.g. `git describe`, or resolve `self.rev` against the repo's tags, or read a checked-in `VERSION`. Then `nix run …/0.1.21#with-mcp -- --version` reports `0.1.21`. Matters for provenance and `--version` output when pinning camas through the flake (and for `camas mcp`'s own reported server version).

> [!WARNING]
> **LLM Disclosure**
>
> This issue was authored by `claude-opus-4-8` on behalf of @JPHutchins, who asked me to pin camas to 0.1.21 via the flake; the built binary reporting `0.0.0+09f0fca` is what surfaced this.

### LINKED ISSUE #172: Scoping: a run-if-changed trigger for whole-project commands, and the full-run () callable footgun
Two related rough edges in the scope system, found while building a mixed **C + Python** gate (`tasks.py` composing ARM/QEMU builds, `nix flake check`, ruff/mypy/pyright, and jphfmt).

## 1. No way to prune a whole-project command in a scoped gate

`scope_to_changed` prunes only `{paths}` leaves; a command with no `{paths}` "always runs" (correct, conservative). But whole-project build/test commands **can't** take `{paths}` — `cmake --build build`, `ctest`, `nix flake check`, `cargo build` would get file args spliced in and break. So in the agent's scoped gate, editing a single `.py` still runs the full C build / QEMU / analyzer, because those tasks have no `{paths}` to prune on.

**Idea:** a scope *predicate* distinct from `{paths}` injection — e.g. `Task(..., when=<prefix | glob | callable>)` — that decides whether the leaf runs for a given changed set **without** substituting the paths into the command. `{paths}` = "narrow this tool to these files"; `when=` = "run this whole-project task only if the change is relevant." That lets a mixed repo keep a comprehensive gate that stays cheap per-edit.

Related: #115 (path-scoped fixer subagents).

## 2. `paths` callable receives `()` on a full run — footgun for stdin tools

On a full run the `paths` callable is invoked with `()`. An extension filter like `lambda c: tuple(p for p in c if p.endswith((".c", ".h")))` then returns `()`, `{paths}` is stripped, and the tool runs with **no file arguments**. For a formatter that reads stdin when given none (jphfmt, `clang-format -`), that's a hang / wrong behavior rather than "format everything." I had to special-case the empty tuple and glob the C tree myself.

**Idea:** either pass the discovered project file set on a full run instead of `()`, or document prominently that a callable must supply its own full-run default, or ship a helper like `by_suffix((".c", ".h"), tree=(...))` that returns a well-behaved scope for both scoped and full runs.

> [!WARNING]
> **LLM Disclosure**
>
> This issue was authored by `claude-opus-4-8` on behalf of @JPHutchins, who asked me to make `tasks.py` the SSOT gate for a mixed C/Python repo and file the friction; both points are things I worked around by hand while getting the agent gate to be both correct and cheap.

### LINKED ISSUE #173: Flake: mypyc packages (default/with-mcp) fail to build when the nixpkgs input is overridden via follows
## Context

Consuming camas via its flake in a Rust/crane project (jphfmt), I added it as an input with the usual `inputs.nixpkgs.follows = "nixpkgs"` (nixpkgs-unstable, rev `d333699`, 2026-07-02) — standard practice to avoid pulling a second nixpkgs — and put `camas.packages.${system}.default` on the dev shell.

## Observed

`nix develop` (which builds the camas package) fails at `pythonImportsCheck`:

```
ModuleNotFoundError: No module named 'camas.main.check'; 'camas.main' is not a package
  File "src/camas/main/entrypoint.py", line 11, in <module>
  File "src/camas/main/dispatch.py", line 42, in <module>
  File "src/camas/main/parser.py", line 23, in <module>
python-imports-check-hook.sh: line 6: … Segmentation fault (core dumped)
  ( … /nix/store/…-python3-3.13.13/bin/python3.13 -c 'import importlib; list(map(...))' )
```

Same for `#with-mcp` and the other `with-*` (mypyc) packages. `#interpreted` (`withMypyC = false`) builds fine under the same overridden nixpkgs.

## Isolation — it's the overridden nixpkgs

With camas's **own** locked nixpkgs (no `follows`), both build cleanly:

```console
$ nix build github:JPHutchins/camas#default    # exit 0
$ nix build github:JPHutchins/camas#with-mcp    # exit 0
```

So the mypyc build isn't broken on `main` — it breaks only when a consumer overrides `nixpkgs`. Note the import check runs under **python3.13** while camas's `requires-python` is `>=3.14`; the foreign nixpkgs appears to shift the interpreter and/or the mypyc-compiled `camas.main` package layout so `import camas.main.check` fails, then the check process segfaults.

## Impact

Following one's own nixpkgs is the norm (avoids a duplicate nixpkgs in the lock/closure and keeps one eval). A consumer doing that can't build the compiled packages at all — they must either drop `follows` (duplicate nixpkgs) or fall back to `#interpreted`, which carries no extras (so no `mcp`). jphfmt went with `#interpreted` + `follows`.

## Suggestions

1. Pin the mypyc build's Python explicitly in `nix/package.nix` (don't inherit the consumer's default `python3`), or otherwise make the compiled package importable regardless of the interpreter the consumer's nixpkgs provides.
2. Failing that, document that `nixpkgs` must not be overridden for the compiled packages, and point consumers who follow their own nixpkgs at `#interpreted`.
3. The segfault in `pythonImportsCheck` (rather than a clean, actionable error) is itself worth hardening.

## Related

- #169 — `with-mcp` runtime crash (PYTHONPATH → `pydantic_core`), built with camas's own nixpkgs; different phase (runtime vs build), same package family.
- #171 — flake version string for tagged builds.
- #134 / #135 — earlier mypyc / Nix `resolveExtra` packaging work.

> [!WARNING]
> **LLM Disclosure**
>
> This issue was authored by `claude-opus-4-8` on behalf of @JPHutchins, who asked me to put jphfmt's environment under Nix (crane) with camas as the task runner on the dev-shell PATH, and to file the friction I hit. Falling back to `#interpreted` is what surfaced this; I isolated it to the `nixpkgs` override by building with camas's own pin.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-fable-5 on behalf of @JPHutchins, who asked me to assess the six field reports (#167, #169–#173) from real-project camas usage and resolve as many as possible in one PR, with coverage for each fix and no regressions.

Closes #167, closes #169, closes #170, closes #171, closes #172, closes #173. #167's two remaining themes are split out to #175 (composite args passthrough) and #176 (diff-aware failure output).

## #172 — `when=` run-if-changed predicate, `by_suffix`, scope authoring warnings

- `Task`/`Group` gain `when=` — a directory-prefix string, a **tuple of prefixes (OR)**, or a `(changed) -> bool` callable. On a scoped run, a leaf whose `when=` doesn't match the changed set is pruned — so whole-project commands that can't take `{paths}` (`cargo build`, `nix flake check`, `ctest`) finally participate in scoping. A full run never consults `when=`. Group `when=` inherits into leaves via `expand_matrix` exactly like `paths`/`env`/`cwd`, with `{VAR}` matrix substitution in the string/tuple forms, and round-trips through the expression/TOML surface (`eval_when`/`to_expression`).
  - The tuple-OR form deliberately diverges from `paths` (single prefix or callable): `paths` must decide *which files inject*, so a tuple is ambiguous there; `when` is a pure boolean, so OR-ing prefixes is the natural mixed-repo idiom.
- `by_suffix(suffixes, default=("."))` — new headline export on `camas`/`camas.v0` — is the safe `PathScope` factory: filters changed files by suffix on a scoped run, returns `default` on a full run, so a stdin-reading formatter never sees an empty `{paths}` (the footgun in the report).
- `scope_warnings` (advisory, never fail anything): `inert_paths` — a leaf's **own** `paths=` on a command with no `{paths}` token (the exact silent no-op the jphfmt follow-up hit; the message points at `when=`); `empty_full_run_callable` — a `{paths}` callable that returns `()` for a full run. Surfaced by `camas --check` on stderr and by `camas_check` (`CheckResponse.warnings` + a rendered advisory block). Group-level `paths=` (a legitimate descendant default) does **not** warn; detection runs on the raw pre-bake tree where own-vs-inherited is still distinguishable.

## #170 — pinned `camas mcp init` launcher + `--launcher`

- The `uvx` fallback now pins to the **running camas release** (`uvx 'camas[mcp]==X.Y.Z'`). A dev/local version (`0.1.22.dev3+g…`) isn't on PyPI to pin against, so it stays unpinned. A PEP 723 requirement in `tasks.py` remains authoritative when present.
- New `camas mcp init [--claude] --launcher uv|uvx|camas` forces the launch strategy — `--launcher camas` is the nix/flake answer (PATH-provided binary, no second PyPI fetch). Each choice errors with one actionable line when its prerequisite is missing. The autofix hook uses the same resolution.
- README updated to match what actually lands (the report caught it overclaiming).

## #169 — PYTHONPATH: fixed at the source, hardened at the wrapper, diagnosable at runtime

Verified root cause: `with-mcp` propagated `python3` + its deps, so a consumer `pkgs.mkShell { packages = [ camas.with-mcp ]; }` had camas's **entire python closure stuffed into `PYTHONPATH`** by the python setup hook — which is also what shadowed the separately-launched `uvx` server's `pydantic_core`.

- `nix/package.nix`: `propagated-build-inputs` emptied (an application is a leaf; its deps are baked into the wrapper) — a consumer mkShell now gets an **empty** `PYTHONPATH` with camas still on PATH. Plus `makeWrapperArgs = [ "--unset PYTHONPATH" ]` so the binary itself survives poisoning from any other source (verified: runs correctly under `PYTHONPATH=/nonsense:/tmp`).
  - Tradeoff, flagged for review: `--unset` also strips `PYTHONPATH` from task subprocesses camas spawns. For a nix-provided camas the ambient value is essentially always leakage, but a deliberately-exported user `PYTHONPATH` won't reach tasks through the nix binary. The `-E` alternative has no clean `buildPythonApplication` knob.
- Diagnosability: a broken `camas[mcp]` dependency import now prints one actionable stderr line — failing module, interpreter, and (when `PYTHONPATH` is set) the leaked-mkShell hint with `unset PYTHONPATH` — and exits 2, instead of the raw traceback Claude Code swallows into "no tools, no error". End-to-end test reproduces the field failure with a shadowed `pydantic_core`.

## #171 — flake version from a CI-gated `VERSION` file

Flake eval genuinely cannot see git tags (`nix flake metadata` confirms only `rev` reaches `outputs`). The tag stays SSOT via enforcement rather than duplication:

- Checked-in `VERSION`; the flake reports it bare on a clean checkout (so `nix run github:JPHutchins/camas/0.1.21#with-mcp -- --version` → `camas 0.1.21`) and `VERSION+<rev>` on a dirty tree — never `0.0.0` again.
- New `version-gate` CI job (tag pushes only, in `publish`'s `needs`) asserts `VERSION == tag`, so a drifted `VERSION` can never publish.
- Known wart: a clean checkout *ahead* of the tag also reports the bare version — flake eval can't distinguish it from the tagged fetch. The always-`+rev` alternative was rejected because the report explicitly wants exact `X.Y.Z` at a tag.
- `.github/scripts/release.py` + a `release` task (`camas release -- 0.1.22`, dogfooding leaf-task `--` passthrough): asserts clean/synced `main`, rejects re-releases and existing tags, bumps `VERSION`, commits `release: X.Y.Z`, tags; pushing stays manual since the tag push is the publish trigger. 11 tests against a scratch repo with a local bare origin. Documented in CONTRIBUTING.

## #173 — mypyc build broken under a consumer-overridden nixpkgs

Verified root cause (not the interpreter-pinning the report guessed): the mypycify glob compiled `camas/main/__init__.py`, and a **compiled** package `__init__` executes its body inside `create_module`, *before* the import machinery sets `__path__` — so its eager import chain hit `from .check import …` while `camas.main` "is not a package". mypy 1.19 (camas's locked nixpkgs) tolerated the ordering; 1.20 (consumer's nixpkgs) doesn't. Fix: exclude `__init__.py` from compilation (it's a one-line re-export). Verified: `nix build .#default --override-input nixpkgs github:nixos/nixpkgs/d333699…` — the exact reported reproduction — now builds, and the locked build still passes.

## #167 — run-log rotation (theme 3)

`.camas/runs/<task>/` now keeps the newest `RUN_LOG_KEEP = 10` run dirs per task — recency by mtime with a sequence-number tie-break (the per-session counter restarts at 1, so numeric order alone lies across sessions; the tie-break also makes pruning deterministic on coarse-mtime filesystems like the nix sandbox). Non-numeric entries are never touched; deletion is best-effort and can never fail a run.

The other two themes are deliberately out of scope here and now tracked on their own: #175 (args through a matrix `Parallel` — the composite-args design question, related to #4) and #176 (diff-aware truncation of huge assertion output). With those split out, this PR closes #167.

## Verification

- `camas check` (format, lint, mypy, pyright, ty, zuban, pyrefly, pytest with doctests): **8/8 green**.
- `camas matrix` (full check across every supported Python from `.python-version`): **green**.
- Coverage extended for every fix: ~60 new tests (scope/matrix/expression/check for `when=`/`by_suffix`/warnings; scaffold/cli/serve for pin/launcher/diagnostic/rotation; wire/result for the warnings channel; release-script suite; inert-paths fixture).
- Nix, empirically: locked `#with-mcp` builds with the full test suite in-sandbox; wrapper contains `unset PYTHONPATH`; `propagated-build-inputs` empty; binary works under a poisoned `PYTHONPATH`; `#default` builds under the consumer's exact overridden nixpkgs rev.
- An Opus docs-consistency audit swept every doc surface (README, tutorial docstring served by `camas_docs`, MCP tool/wire descriptions, CLI help, agent/skill templates, starter, nix comments, CONTRIBUTING); its findings (one stale README claim, missing `when=`/`by_suffix`/release docs) are fixed in this PR.

## API notes for review

- `by_suffix` widens the headline namespace (`tests/test_v0.py` `HEADLINE` updated accordingly) — trivial to demote to a submodule export if preferred.
- `CheckResponse` gains `warnings: tuple[str, ...] = ()` (breaking-OK per alpha v0).

https://claude.ai/code/session_01FZpvKPhRH1QhKYZcQhRS2D


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by claude-fable-5 on behalf of @JPHutchins, who asked me to verify the DeepSeek V4 review findings above and apply the ones that hold up.

Verified every finding against the code; 76f2911 applies the confirmed ones. Per-finding verdicts:

| # | Finding | Verdict | Action |
|---|---------|---------|--------|
| 1 | matrix wrappers drop `paths`/`when` vs. docstring claim | **Confirmed (docstring)** — the wrappers never retained `paths` even before this PR; baking into leaves is the design. | Docstring now states the actual contract: non-matrix rebuilds retain group values; matrix wrapper nodes carry env/cwd only. |
| 2 | callable `when` in `__hash__` breaks dedup | **Refuted as a defect** — identity hashing for callables is the established `paths` semantics; two distinct lambdas are legitimately distinct scopes, not equal tasks. | None. |
| 3 | `empty_full_run_callable` warns on explicit `by_suffix(..., default=())` | **Acknowledged, by design** — the warning is advisory and truthful (the full-run strip is real); an explicit `default=()` is a conscious choice to ignore it. | None. |
| 4 | `scope_warnings` invokes the `paths` callable at check time | **Refuted** — `with_default_paths` already calls it with `()` before every full run; check mirrors the documented PathScope contract exactly, no new side-effect channel. | None. |
| 5 | `when` gate vs `{paths}` scope backslash asymmetry | **Confirmed** — real latent Windows-shaped inconsistency (unreachable via `to_changed`, which normalizes, but the in-function defense was asymmetric). | `_resolve_leaf` normalizes once; both consumers see the same POSIX set. Regression test added. |
| 6 | `isdigit()` admits `'²'`, `int()` raises | **Confirmed** — would crash best-effort pruning on a foreign dir. | `isdecimal()` (exactly the `int()`-safe predicate). Regression test added. |
| 7 | duplicated warning-dedup logic | **Confirmed** — a known artifact of parallel implementation scoping. | Shared `scope_warning_messages` in `core/scope.py`; both surfaces consume it. |
| 8 | hand-rolled `parse_launcher` vs local-argparse pattern | **Confirmed** — `fix_cli`/`parse_gate_args` are the precedent. | Rewritten with `argparse` (`choices=`, `exit_on_error=False`), same `ValueError` contract. |

Verified after the changes: `camas check` 8/8 green, coverage gate 100%.

https://claude.ai/code/session_01FZpvKPhRH1QhKYZcQhRS2D

PR #181, opened 2026-07-06
feat: dogfood-friction batch — run default, version-skew hints, github_default docs (#177)

### LINKED ISSUE #177: Dogfooding friction: MCP version-skew diagnostics, default-task/github_default ergonomics (from code-review build)
> [!WARNING]
> **LLM Disclosure**
>
> This issue was authored by claude-fable-5 on behalf of @JPHutchins, who asked me to log any camas MCP / plugin / fixer friction encountered while building the `code-review` project and file it upstream once the work completed. These are field notes from a multi-agent session (one orchestrator + ~10 opus/sonnet subagents) that ran camas as its gate throughout.

## Context

Dogfooding camas 0.1.21 → 0.1.22 as the CI/gate SSOT for the `code-review` repo, via the MCP server (`uv run tasks.py mcp`), the Claude Code gate skill, and `camas-fixer` subagents. Roughly a dozen agents exercised `camas_check`/`camas_run`/`camas_gate`/`camas_fix` and the `uv run tasks.py` CLI. Most of it worked cleanly — `when=` (new in 0.1.22) and the fixer flow in particular. The items below are the friction, ordered by impact, each with a concrete repro and a one-line suggestion. Nothing here is blocking; several are papercuts that repeatedly cost agents a re-diagnosis round.

---

### F1 (HIGH, hit by 4+ agents) — MCP server / `tasks.py` version skew is misdiagnosed

The MCP server starts once per session via `uv run tasks.py mcp`. When `tasks.py`'s PEP-723 camas pin was bumped mid-session (0.1.21 → 0.1.22) and the taskfile started using a 0.1.22-only feature (`when=`), every camas MCP tool began returning:

```
error: /home/.../tasks.py:22
    actionlint = Task(
                 ^
TypeError: Task.__init__() got an unexpected keyword argument 'when'
```

The diagnostic points at `tasks.py` as if the taskfile were malformed, when the real fault is the **running server process** being an older camas than the file now pins. The orchestrator burned a debugging round (checked PyPI upload timestamps, tag list, `camas_docs` source path) to land on "stale server — reconnect via `/mcp`," and several subagents independently re-derived the same conclusion (one close-but-guessing: "an unrelated version mismatch between the installed MCP server and the project's pinned camas").

**Suggestion:** on a load error, have the server compare its own running `camas.__version__` against the PEP-723 pin in the tasks file and, on mismatch, prepend one actionable line — e.g. *"MCP server is running camas 0.1.21 but tasks.py pins 0.1.22 — reconnect the MCP server to pick up the new version."* Bonus: include the running server version in every `camas_check` payload.

### F2 (MEDIUM, consequence of F1) — load error doesn't hint the CLI fallback

While the server was stale, every subagent that hit F1 had to be told (or work out) that `uv run tasks.py <task>` still works, because the CLI resolves the pin fresh per invocation. The load-error response could itself suggest the CLI fallback so agents self-recover without orchestrator hand-holding.

### F3 (LOW) — `camas_run` with no args has no default-task behavior, unlike the bare CLI

`camas_run` with no arguments returns `camas_run requires 'task' (or pass 'under' to budget the default task)`, whereas bare `camas` runs the configured `default_task`. The asymmetry cost one round-trip. Either honor the default task, or name it in the error (*"default task is 'ci'"*).

### F6 (LOW) — `camas_list` reports `github_default: null` even with a CI workflow present

`camas_list` returns `github_default: null` / `is_github_default: false` for every task, though `.github/workflows/ci.yaml` exists and the field's docs describe it as "the exact task this project's CI runs." An agent trusting that field to locate the CI-equivalent task gets nothing and must fall back to `is_default` (`ci`). If the workflow→task mapping can't be inferred, one doc line (*"github_default is null unless the workflow invokes `camas run <task>` explicitly"*) would save a guess.

### F7 (LOW / informational) — concurrent leaves under `Parallel` can flake a tight per-test timeout

A `camas_run task="ci"` reported 2/283 test failures, both `Test timed out in 5000ms` on trivially-synchronous tests, in the two heaviest vitest files; an immediate isolated `npx vitest run` passed 283/283 in 1.4s. Root cause is CPU contention — `ci = Parallel(checks, build)` runs vitest (itself internally parallel) alongside `build`/`lint`/`typecheck`, and under load vitest's first ajv compile blew the 5s per-test default. **This is primarily a consumer fix** (raise `testTimeout`), not a camas defect — but a green-in-isolation suite flaking red under `Parallel` reads to an agent as a spurious CI failure and cost one re-run to rule out a regression.

**Suggestion (ergonomics only):** in the run summary, note when a failing leaf's wall-clock overlapped N other running leaves (a *"ran under contention with: build, lint"* hint), so an agent can distinguish a load-induced flake from a real failure.

---

### Positives (worth keeping)

- **F4 — `when=` (0.1.22) landed exactly right.** The tuple-of-prefixes `when=` scoping was the correct shape for a no-`{paths}` lint leaf (`actionlint` over two workflow dirs) on the first try, and `camas --check` passing with zero `scope_warnings` gave immediate confidence. Python-side globbing in `tasks.py` was easy once the tutorial's "tasks.py is real Python — source values from the project" idiom clicked. The documented idiom carries real weight for agents.
- **The gate + `camas-fixer` flow was frictionless once the server was healthy.** After the F1 reconnect, five separate implementer/reviewer agents drove `camas_gate`/`camas_fix`/`camas_run` and a background `camas-fixer` to green with zero residuals and no complaints; `camas_run` reproduced CI exactly every time. Three of the last four agents reported no friction at all.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-fable-5 on behalf of @JPHutchins. She asked me to drive resolution of #177 end to end — branch, implementation, and PR — coordinating opus planner/reviewer agents over sonnet implementers, and to exercise the new review.yaml flow on the result.

Resolves the approved items from the code-review dogfooding field report (#177).

## F3 — `camas_run` default task (high priority)

- New `Claude(default=...)` field; `Config.run_default()` resolves the chain **`Claude.default` → `Claude.check` → `Config.github_task` → `Config.default_task`** (implemented as `agent.default` else `gate_check(github=True)`, so the chain is exactly the approved order with no duplicated fallback logic).
- All three no-task paths are unified on that resolver: plain `camas_run`, `dry_run=true`, and `under=` budgeting. Previously they disagreed (dry_run used `gate_check(github=False)`, budget used `default_task` only, plain run just errored).
- `matrix_overrides` and `args` now apply to the resolved default exactly as they would to a named task (caught by the opus review pass — the first cut dropped them silently on the no-task path).
- When nothing resolves, the error is actionable: `camas_run has no task to run: name a task, or set a project default in tasks.py — Config(default_task=…), Config(github_task=…), or Config(agent=Claude(fix=…, default=…)).`
- Gate semantics (`gate_check`/`gate_fix`) are untouched; `under=` with a named task is untouched.

## F1/F2 — load-error version-skew diagnostics

- A tasks-file load error on `camas_list`/`camas_run`/`camas_gate`/`camas_fix` now leads with, when the running server fails the PEP-723 pin:

  > MCP server is running camas 0.1.21 but tasks.py pins camas==0.1.22 — reconnect the MCP server to pick up the new version.
  >
  > Run the task directly with `uv run tasks.py <task>` in the meantime: it resolves the pinned camas version fresh per invocation, so it works while the server is stale.

  followed by the existing `Tasks unavailable — …` text. Non-skew load errors are unchanged (the CLI fallback would fail identically on a genuinely broken file, so it isn't suggested there).
- One deliberate deviation from the issue's literal wording: the pin renders as `camas==0.1.22` rather than bare `0.1.22`, so non-`==` specifiers (`>=`, `<`) read accurately.
- The headless `camas mcp gate` load error gets the same skew + `uv run tasks.py` note (reconnect advice would be meaningless for a fresh subprocess).
- Internals: `check_version_pin` is refactored to delegate to a structural `version_skew() -> VersionSkew | None`; its success-path warning string is byte-identical to before.
- Bonus (approved): every `camas_check` payload now carries `server_version` (`null` when distribution metadata is unavailable), and `check_text` renders it as `camas server version: X`.

## F6 — `github_default` is declarative (docs only)

All four surfaces now state that `github_default` is declared via `Config(github_task=...)`, is `null` when unset, and is never inferred from CI workflow files: the `camas_docs` tutorial (`__init__` docstring), README, the `camas_list` tool description, and a schema-visible `Field(description=...)` on `ListResponse.github_default`.

## Out of scope

- **F7** (contention hint on Parallel leaves): dropped as consumer-side — the fix is raising the test timeout or isolating working dirs, not a camas feature.
- **F4/F5 positives**: nothing to do.

## Verification

- Full local gate green twice (post-implementation and post-review-fix): ruff fix/format, mypy + pyright + ty + zuban + pyrefly, coverage at 100% branch (`uv run camas`).
- 1300+ tests pass including ~20 new ones covering every new branch: the full `run_default` precedence chain, promotion of the new field, no-task run/dry_run/under behavior (named, anonymous, absent defaults; overrides; unknown axis; args), skew vs no-skew load errors on all four tools plus gate CLI, `server_version` presence/degradation, and doc-surface guards (including that the `github_default` description reaches `model_json_schema()`).

### DIFF
--- a/.github/workflows/claude-code-harness.yaml
+++ b/.github/workflows/claude-code-harness.yaml
@@ -9,7 +9,7 @@ name: Claude Code hook harness
 # PostToolBatch hook fires, and the fixer reaches green via the MCP camas_gate tool.
 #
 # It uses the same DeepSeek backend the agentic review workflow does (ANTHROPIC_BASE_URL +
-# ANTHROPIC_AUTH_TOKEN=DEEPSEEK_API_KEY, a burner key with a hard spend cap), on the cheap
+# ANTHROPIC_AUTH_TOKEN=MODEL_API_KEY, a burner key with a hard spend cap), on the cheap
 # deepseek-v4-flash tier. The schedule catches a Claude Code release changing the contract even
 # when camas itself has not changed.
 on:
@@ -74,12 +74,12 @@ jobs:
           CAMAS_CC_E2E: "1"
           CAMAS_CC_MODEL: deepseek-v4-flash
           ANTHROPIC_BASE_URL: https://api.deepseek.com/anthropic
-          ANTHROPIC_AUTH_TOKEN: ${{ secrets.DEEPSEEK_API_KEY }}
+          ANTHROPIC_AUTH_TOKEN: ${{ secrets.MODEL_API_KEY }}
           # The camas-fixer subagent is model: haiku; point the haiku tier at DeepSeek so the
           # fixer-loop test (criterion #3) drives a real subagent round-trip, as claude-code-review does.
           ANTHROPIC_DEFAULT_HAIKU_MODEL: deepseek-v4-flash
           CLAUDE_CODE_SUBAGENT_MODEL: deepseek-v4-flash
           UV_PYTHON_DOWNLOADS: never
         run: |
-          [ -n "$ANTHROPIC_AUTH_TOKEN" ] || { echo "::error::DEEPSEEK_API_KEY secret is not set"; exit 1; }
+          [ -n "$ANTHROPIC_AUTH_TOKEN" ] || { echo "::error::MODEL_API_KEY secret is not set"; exit 1; }
           uv run --python 3.12 pytest tests/claude_code/ -v --no-cov -p no:cacheprovider
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ Because `github_task` reproduces CI, running it before a push catches a CI failu
 echo 'exec camas check' > .git/hooks/pre-push && chmod +x .git/hooks/pre-push
 ```
 
-`git push --no-verify` still bypasses it deliberately. An LLM agent needs no hook — `camas_list` reports the CI task as `github_default`, so it runs `camas_run` with that name before pushing.
+`git push --no-verify` still bypasses it deliberately. An LLM agent needs no hook — when a `github_task` is declared, `camas_list` reports it as `github_default`, so the agent runs `camas_run` with that name before pushing. The field is `null` when no `github_task` is set; camas never infers it from your CI workflow files.
 
 ## Time budget (`--under`)
 
--- a/src/camas/__init__.py
+++ b/src/camas/__init__.py
@@ -53,34 +53,37 @@ real ordering (a mutating step, or one that consumes a prior's output).
 (``github_task`` takes over under GitHub Actions) — so a CI step is just
 ``camas`` (or ``uv run camas``), which won't go stale if you change
 ``github_task``. That same ``github_task`` reproduces CI locally — run it
-before you push to catch a CI failure first: ``camas_list`` surfaces it as
-``github_default`` for an agent to ``camas_run``, and a named CI task
-doubles as a one-line git ``pre-push`` hook.
+before you push to catch a CI failure first: ``camas_list`` surfaces the
+declared ``github_task`` as ``github_default`` for an agent to ``camas_run``
+(``null`` when no ``github_task`` is set — camas never infers it from CI
+workflow files), and a named CI task doubles as a one-line git ``pre-push``
+hook.
 
 These five are the unversioned alias for the latest API generation; to
 pin a generation, import from its namespace (``camas.v0``). See the
 README's Versioning section.
 
-``Config(agent=Claude(fix=..., check=...))`` wires the Claude Code
-plugin's gate. ``fix`` is the deterministic, behavior-preserving autofix
-node the ``PostToolBatch`` hook runs over the changed files (scope it with
-``{paths}``, on a leaf or a whole group; zero model tokens) — declared, not inferred from
-``mutates=``, since a mutating leaf may be a compiler or codegen rather
-than a fixer. A leaf whose command can't take ``{paths}`` (``cargo build``,
-``nix flake check``) instead sets ``when=`` — a directory-prefix string, a
-tuple of prefixes, or a ``(changed) -> bool`` callable — so the gate skips
-it on a scoped run the change doesn't touch, and never consults it on a
-full run; ``by_suffix`` builds a suffix-filtering ``paths`` scope for a
-``{paths}`` leaf that should still run against the whole project on a
-full run. ``check`` is the node the gate validates (``None`` defers
-to ``default_task``/``github_task``), scoped by ``--paths`` and
-time-boxed by ``--under``; the plugin delegates it to a background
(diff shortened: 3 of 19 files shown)


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by claude-fable-5 on behalf of @JPHutchins — the verification pass over the automated review above, per this repo's practice of confirming or refuting every bot finding against the code.

## Review-finding verdicts

| # | Finding | Severity | Verdict | Disposition |
|---|---|---|---|---|
| 1 | `camas_list` reports `default: null` when only `Claude(default=...)` is configured (`catalog.py:35`) | major | **CONFIRMED** | **Fixed** — additively: `ListResponse` gains `run_default` (what a no-task `camas_run` resolves), with a `no-task camas_run runs: <name>` text line when it differs from the developer default, schema descriptions, and tests. `default` deliberately stays `Config(default_task=...)`: repointing it at `run_default()` would make `[default]` misstate what the bare CLI runs and collapse into `[github]` whenever `github_task` is set. |
| 2 | `check_call` overwrites `session.project` without re-pinning `session.version_warning` | minor | **CONFIRMED** | **Fixed** — `session.refresh()` (which the docstring already promised: "re-resolve and re-pin"). The warning test now exercises a real PEP-723 pin mismatch instead of a hand-set warning, plus a stale-warning-cleared regression test. |
| 3 | `init_call` bypasses `session.refresh()` — same defect | minor | **CONFIRMED** | **Fixed** — same one-line change, with a re-pin regression test. |
| 4 | `gate_cli_load_error` duplicates version-skew formatting instead of delegating to `format_version_skew_hint` | nit | **REFUTED** | Deliberate divergence, not drift: the MCP hint's advice ("reconnect the MCP server") is wrong for the headless gate, which is a fresh subprocess per run — its correct advice is the `uv run tasks.py` fallback alone. The two messages share only the factual skew clause; extracting a shared helper would couple texts that must differ. |

Fix commit `76c53d7` is ready locally (full gate green: ruff, mypy/pyright/ty/zuban/pyrefly, coverage 100% branch) and will be pushed once the in-flight review.yaml updates land on main, to avoid burning a review round mid-upgrade.

**Note on "see inline comments for details":** no inline comments exist on this PR — the poster's same-SHA dedup matched the earlier gate-skip placeholder sticky and suppressed the inline pass. Diagnosed and filed upstream as [code-review#5](https://github.com/JPHutchins/code-review/issues/5).

PR #190, opened 2026-07-08
feat: monorepo support — compose child tasks.py via explicit Project() references

### LINKED ISSUE #189: Initial monorepo support
Composbale by file structure

Design sketch:

- tasks.py will DFS its child folders finding other tasks.py
- those tasks.py are added and namespaced to the main tasks.py
- tasks.py IDs modules, sorta like __init__.py, but it doesn't require continuity

Examples
```
tasks.py
libs/search/tasks.py
libs/history/tasks.py

finds search and history namespaces

tasks.py
libs/tasks.py
libs/search/tasks.py
libs/history/tasks.py

finds libs, libs.search, and libs.history namespaces

tasks.py
akld/fjklef/slfiehf/libs/search/tasks.py

finds search namespace (nothing in between)

```

This does risk namespace collision, but that can be resolved when it occurs.

Rquirements:
- each tasks.py is still runnable from where it lives
- when an imported namespaced task is run from a parent tasks.py, every node gets the pwd of the leaf tasks.py as cwd. If it already has cwd, that means the pwd gets put in front to resolve relative paths (or ignored if the node dfines and aboslute cwd (unlikely))

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-4-8 on behalf of @JPHutchins. JP handed off this PR (issue #189, then CI-failing) for review and finishing; mid-review she redesigned the feature from automatic file-structure discovery to explicit `Project()` references — because composing children into a single parallel CI task needs a *typed, named handle* that discovery could never provide. Opus designed the type approach with her (mirroring the existing `Ref` sentinel) and implemented it.

Closes #189.

## What

A `tasks.py` composes others by **importing** them. `libs = Project("libs")` binds another `tasks.py` (a directory resolves its `tasks.py`) as a task node — a self-contained child project — that runs whatever a bare `camas` runs in that directory, and mounts the child's own tasks under the binding name:

```python
from camas import Config, Parallel, Project

libs = Project("libs")            # camas libs, camas libs.search.lint, ...
api  = Project("services/api")    # name the handle whatever you like; gaps are just paths

_ = Config(
    default_task=...,                     # what bare `camas` runs here
    github_task=Parallel(libs, api),      # each child's CI default, composed in parallel
)
```

- **Context resolution.** A reference resolves to the child's default locally, its `github_task` under CI (`GITHUB_ACTIONS`), its agent default under an agent — so a `Parallel` of references *is* the composite CI task, each child contributing its own. Resolved at load time (load and dispatch share the invocation's env, so `--list` shows exactly what will run in the current context).
- **Namespacing.** `camas libs.search.lint` reaches a task the child exposes, because `libs/tasks.py` itself did `search = Project("search")`. Expressions compose across namespaces (`camas '{libs.search.lint, api.deploy}'`). Each `tasks.py` stays independently runnable — the nearest one up from cwd loads.
- **Anchoring (fixes a field wart).** A node's `cwd` and its `paths=`/`when=` scopes are authored relative to its own `tasks.py`, rebased across the boundary and resolved at spawn via `run(base=)`, so invoking `camas` from a subdirectory no longer misresolves relative cwds. Scopes (strings, tuples, callables) rebase so `--paths`, the gate, and the fixer stay correct on composed trees; callables receive the child-relative changed set.
- **Private, immutable child.** Nothing crosses the boundary except the resolved node — a child's `Effect` classes, its matrix, and `--` passthrough stay local. Children are referenced by path within the importing file's directory. A broken child attributes the load error to the child file (`--list`/`--tree` keep working); the MCP server sees the same composed view.

## Type approach

`Project` mirrors the existing `Ref` parser sentinel: `ProjectRef` is **not** in the `TaskNode` union — `Project()` rides it as a `TaskNode` via one boundary `cast` (as `children()` already does for `Ref`), and the loader resolves it into concrete nodes before the engine ever runs. So `Parallel(libs, api)` type-checks and **the engine is untouched** — no new `match` arm anywhere.

## What this removes

The whole discovery engine: the filesystem walk, the camas-import sniff, the collision auto-extension, and the `discover`/`discoverable`/`ignore_dirs` config knobs. Collisions are now impossible — the namespace segment *is* the Python variable name. The cross-boundary `cwd`/`paths`/`when` rebasing (the genuinely hard part) is kept. Net **−506 LOC**.

## Fixture as UX vehicle

`tests/fixtures/monorepo/` demonstrates the permutations in one small tree — namespace mounting, two-level `cwd` rebasing, a gap dir named by explicit import (`services/api`), `paths=` rebasing (`web`), context resolution (`web`'s default vs github), and the `Parallel` github composite — with `monorepo-badchild/` holding the child-attribution case. Leaves print their effective cwd, so composition is directly observable: edit the fixture and re-run `tests/test_monorepo_fixture.py`.

## Verification

`camas check` green (ruff format/lint, mypy, pyright, ty, zuban, pyrefly, full pytest incl. doctests) and `camas coverage` at the enforced 100% branch threshold — **and** a local `CAMAS_USE_MYPYC=1` wheel passing the full not-slow suite (1243 passed) with the loader interpreted and the hot modules compiled (mypyc's mypy is stricter than the interactive one, so the wheel build is the real check for the `wheels`/`nix` jobs).

The loader (renamed from `discover.py`) is a `compose/` package — `errors.py` (`ProjectLoadError`), `rebase.py` (the `cwd`/`paths`/`when` re-anchoring), `scope.py` (composition + the `load_scope`/`load_py_tasks_state`/`state_from_scope` seams). It stays interpreted under mypyc for free: the non-recursive `src/camas/main/*.py` glob never reaches a subpackage, and `ProjectLoadError` (caught across compiled call boundaries) must not be a native class.

https://claude.ai/code/session_01JaG3b7vEC2RUyYYbmKbTix


src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+ancestor.
+
+Rebasing (:func:`rebase_cwd`, :func:`rebase_paths`, :func:`rebase_when`, :func:`rebase_tree`)
+re-anchors a child's ``cwd``/``paths``/``when`` into the composing directory's frame, the same
+way :mod:`camas.core.scope` treats a leaf's own scope versus one inherited from an enclosing
+group — a leaf that sets none of these keeps deferring, now one level further out.
+
+:func:`composed_view` and :func:`load_py_state` are the ``tasks.py``-file entry points;
+:func:`state_from_scope` is the sibling for a ``run_cli(globals())`` (PEP 723) scope.
+"""
```
JPHutchins: LLM slop docs, remove

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+
+if TYPE_CHECKING:
+	from collections.abc import Mapping, Sequence
+
+	from ..v0.task import PathScope, TaskNode, WhenPredicate
+	from .state import TasksState
+
+
+IGNORE_DIRS: Final = frozenset({"node_modules", ".venv", "venv", "__pycache__"})
+"""Directory basenames :func:`discover_children` never descends into, beyond hidden ones."""
```
JPHutchins: This needs to be hoisted to Config, the defaults can be there, user can append/override.

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+
+
+def is_pruned_dir(name: str) -> bool:
+	"""True when a directory basename is hidden or a known non-task directory
+	(:data:`IGNORE_DIRS`) that :func:`discover_children` never descends into.
+
+	>>> is_pruned_dir(".git"), is_pruned_dir("node_modules"), is_pruned_dir("src")
+	(True, True, False)
+	"""
+	return name.startswith(".") or name in IGNORE_DIRS
```
JPHutchins: Candidate for LLM slop

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+
+
+def discover_children(root_dir: Path) -> tuple[Path, ...]:
+	"""The nearest descendant directories under ``root_dir`` containing a camas ``tasks.py``
+	(:func:`is_camas_tasks_file`), found by depth-first, name-sorted traversal. Hidden
+	directories, :data:`IGNORE_DIRS`, and symlinked directories are pruned. A directory that
+	itself has a camas ``tasks.py`` is a leaf of the search — its own descendants are never
+	visited, so a nested ``tasks.py`` composes only its own subtree. A directory whose
+	``tasks.py`` is not a camas file is walked like any other, so foreign task modules neither
+	execute nor block discovery beneath them.
```
JPHutchins: Too much uncheckable detail in the comment. The code is the spec, don't violate SSOT. Slop.

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+			(entry,) if is_camas_tasks_file(entry / "tasks.py") else discover_children(entry)
+		)
+	)
+
+
+def rebase_cwd(cwd: Path | None, rel: PurePosixPath, *, is_root: bool) -> Path | None:
+	"""``cwd`` re-anchored into the composing directory's frame at ``rel``: an unset ``cwd``
+	anchors to ``rel`` only at the tree's own root (elsewhere it stays unset, so
+	:func:`camas.core.matrix.expand_matrix` inherits the rebased ancestor's ``cwd`` instead); an
+	absolute ``cwd`` is left as-is; a relative ``cwd`` is nested under ``rel``.
```
JPHutchins: What does this say that the code doesn't say? It's slop.

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+	"""``path`` with the ``rel`` prefix removed, or ``None`` when it doesn't lie under ``rel``
+	(segment-wise, matching :func:`camas.core.scope._within`).
+
+	>>> _strip_rel("services/api/src/a.py", PurePosixPath("services/api"))
+	'src/a.py'
+	>>> _strip_rel("services/other/x", PurePosixPath("services/api")) is None
+	True
+	"""
+	parts = PurePosixPath(path).parts
+	prefix = rel.parts
```
JPHutchins: We don't alias, remove useless alasing.

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+class ComposeChildError(Exception):
+	"""Any exception raised while loading or composing a child ``tasks.py``, carrying its
+	``source`` so the caller can attribute the failure to the child instead of the composing
+	parent's own file.
+	"""
+
+	def __init__(self, source: Path, cause: Exception) -> None:
+		super().__init__(f"{source}: {cause}")
+		self.source = source
+		self.cause = cause
```
JPHutchins: Final, aren't they?

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+	"""``own`` (already loaded from ``root_dir / "tasks.py"``) composed with every discovered
+	child (:func:`discover_children`) it is configured to discover: each child's tasks are
+	rebased (:func:`rebase_tree`) and merged under ``f"{mount}.{name}"`` — ``mount`` the child
+	directory's name, extended with parent directory segments while it collides
+	(:func:`resolve_mounts`) — and its ``Config.default_task``, if any, under the bare
+	``mount``; a child recurses through this same composition before its own tasks are merged,
+	so a grandchild's tasks arrive already namespaced under the child's own mount. A child
+	whose ``Config.discoverable`` is ``False`` (and everything below it) is skipped entirely.
+	Only ``default_task`` is taken from a child's ``Config`` — its ``scope_effects`` are not
+	composed.
```
JPHutchins: Too much detail. The code is the spec, this is garbage.

src/camas/main/discover.py
```diff
@@ -0,0 +1,578 @@
...
+	"""``tasks_py`` loaded and composed with its discovered descendants (:func:`compose_from`)."""
+	return compose_from(tasks_py.parent, load_own(tasks_py))
+
+
+def load_py_state(path: Path) -> TasksState:
+	"""``path`` loaded and composed (:func:`composed_view`) as a
+	:class:`~camas.main.state.TasksState`: any exception during composition becomes a
+	:class:`~camas.main.state.LoadErr` attributed to the file that raised — the composing
+	parent's own file, or, via :class:`ComposeChildError`, the child that failed to load.
+	"""
```
JPHutchins: Doesn't the code say this same thing, more precisely?

README.md
```diff
@@ -253,6 +253,30 @@ For the Claude Code plugin, you **register** the auto-fix node — whatever you
...
+
+```python
+_ = Config(discover=False)      # this file composes nothing below it
+_ = Config(discoverable=False)  # this file (and its whole subtree) is private to ancestors
+```
+
+A colliding namespace extends itself with parent directory segments until unique: `services/api` and `vendor/api` mount as `services.api` and `vendor.api`, and a root `api` binding pushes a discovered `services/api/tasks.py` out to `services.api`. A collision that survives full extension (a root `search` binding beside a `search/` directory) fails the load naming both files, as does a discovered child that raises on evaluation — attributed to the child file, and `--list`/`--tree` still work and point at it. A broken child can't hide behind `discoverable=False`: the flag lives in the file, so the file must evaluate for it to be read. Composition takes tasks only — a child's `Effect` classes and the rest of its `Config` stay local — and `[tool.camas.tasks]` in a `pyproject.toml` never composes, as root or as child. The [monorepo fixture](https://github.com/JPHutchins/camas/tree/main/tests/fixtures/monorepo) exercises every permutation.
+
```
JPHutchins, line 275: Trim this down to about half the size, less prose, more vague, this feature is subject to rapid change.

JPHutchins (commented):
High ratio of LLM comment slop. Comments violate SSOT. Follow repo guidelines.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by claude-opus-4-8 on behalf of @JPHutchins, summarizing how the mid-review redesign to explicit `Project()` references and the CI fix landed, in response to JP's inline review.

Two things happened after your review: the **CI fix** (the `wheels`/`nix` red was a compiled `discover.py` silently skipping discovery + a native `ComposeChildError` breaking `except` matching — resolved by keeping the module interpreted under mypyc), and then the **redesign** you called for — auto-discovery → explicit `Project()` references.

**On your review comments** — most are resolved by deletion, since the discovery engine is gone:

| Comment | Resolution |
|---|---|
| module docstring / `discover_children` / `compose_from` "slop", "too much detail" | module deleted; `compose.py`'s docstrings are vague-WHAT + doctests |
| `IGNORE_DIRS` → hoist to `Config` | gone — no walk, no ignore set |
| `is_pruned_dir` slop candidate | gone |
| `rebase_cwd` "what does this say the code doesn't" | trimmed to one line + doctests |
| `_strip_rel` "we don't alias" | rewritten with `is_relative_to`/`relative_to`, no aliases |
| `ComposeChildError` attrs "Final, aren't they?" | `ProjectLoadError.source`/`.cause` are `Final` |
| `load_py_state` "code says this more precisely" | trimmed |
| README "trim to ~half, vaguer" | halved, vaguer |

**On the design:** `Project` mirrors `Ref` — one boundary `cast`, engine untouched (per your approval of that exception). Context resolves at load time (local / CI / agent), the child is private + immutable, and `--list` shows the context-resolved forms.

**Verified both paths:** `camas check` (all six checkers + doctests) and `camas coverage` at 100% branch, **plus** a local `CAMAS_USE_MYPYC=1` wheel running the full not-slow suite — which caught a `type(group)(...)` narrowing issue that the interpreted `camas check` mypy did not (mypyc's mypy is stricter; the wheel build is the real gate for `wheels`/`nix`). CI is green across all 112 checks. Net **−506 LOC**.

src/camas/core/execution.py
```diff
@@ -13,6 +13,7 @@
 from contextlib import nullcontext
 from dataclasses import dataclass
 from datetime import datetime
+from pathlib import Path  # noqa: TC003
```
JPHutchins: Why noqa? Isn't this correctly saying that this should be in the TYPE_CHECKING block?

src/camas/core/execution.py
```diff
@@ -73,6 +74,7 @@ class RunContext(NamedTuple):
 	limiter: Limiter
 	interrupts: Interrupts
 	states: Sequence[LeafState]
+	base: Path | None
```
JPHutchins, line 77: Might merit a docstring about why the base path could be None. This would be an SSOT that should not be repeated.

base: Path | None
"""docstring"""

src/camas/main/compose.py
```diff
@@ -0,0 +1,427 @@
+# SPDX-License-Identifier: MIT
+# SPDX-FileCopyrightText: 2026 JP Hutchins
+
+"""Load a ``tasks.py`` scope, resolving its ``Project`` references into the composed task tree.
+
+A :func:`camas.v0.task.Project` binds another ``tasks.py`` as a task node; the referenced
+child runs its context default and mounts its own tasks under the binding's name. Its
+``cwd``/``paths``/``when`` are re-anchored (:func:`rebase_tree`) into the referencing
+directory's frame, the same way :mod:`camas.core.scope` treats a scope inherited from an
+enclosing group.
```
JPHutchins: Remove. The source code clearly states all of this.

src/camas/main/compose.py
```diff
@@ -0,0 +1,427 @@
...
+
+	own: Final = LoadOk(
+		tasks=name_scope_bindings(resolved_scope),
+		source=source,
+		scope_effects=name_scope_effects(resolved_scope),
+		config=name_scope_config(resolved_scope),
+	)
+	merged: Final = {**own.tasks, **namespaces}
+	reject_reserved_names(merged)
+	return own._replace(tasks=merged)
```
JPHutchins: This function is too large. Change compose module to a folder and move this into its own file.

src/camas/main/dispatch.py
```diff
@@ -450,18 +439,26 @@ def dispatch(state: TasksState, argv: list[str] | None = None) -> None:
...
 			assert_never(state)
 
 
 def _load_py(path: Path) -> TasksState:
 	"""Evaluate ``path`` and return a :class:`LoadOk` / :class:`LoadErr`."""
-	try:
-		return load_tasks_from_py(path)
-	except Exception as e:
-		return LoadErr(source=path, exception=e)
+	return load_py_state(path)
```
JPHutchins: Maybe change name of func to load_py_tasks_state?

JPHutchins (commented):
Looking pretty good!

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This comment was authored by claude-opus-4-8 on behalf of @JPHutchins, who asked me to implement the review feedback on PR #190 — her inline comments, the Copilot review, and the automated code-review bot's two findings. It reports what I changed and my verdicts on the two bot findings (one fixed, one refuted). Landed in `b61f520`.

## Bot findings — verified

| Finding | Verdict | Action |
|---|---|---|
| 🔴 `serve.py` `gate_cli` uses the wrong base for `to_changed` and omits `base` from `run_gate` | **CONFIRMED** | Fixed + regression test |
| 🔴 `compose.py` inline `ProjectRef` in a `Group` never mounts the child namespace | **REFUTED** (by design) | Documented with a test |

**gate_cli base (confirmed).** The headless path passed `project_base()` (the invocation dir, or `CLAUDE_PROJECT_DIR`) straight through as the rebasing frame and called `run_gate` with no `base`. Invoked from a child cwd, `resolve_project_quiet` walks *up* to the real `tasks.py`, but `--paths` and leaf `cwd`s then rebased against the invocation dir — and leaf `cwd`s fell back to the process cwd. The MCP server path (`gate_for`) already did this right via `base_for`. Fixed by extracting a shared `base_from(source, fallback)` (which `base_for` now also uses) and threading `base=` into `run_gate`. New test `test_gate_cli_anchors_leaf_cwd_to_tasks_dir_from_subdir` runs the gate from a subdirectory and asserts the leaf spawns in the resolved `tasks.py` dir — it fails without the fix.

**inline Project namespace (refuted).** This is by design, not a bug. A namespace segment *is* the Python binding name — that's the whole model. `libs = Project("libs")` mounts `libs.build` because the binding names it; an anonymous `Parallel(Project("libs"))` has no name to mount under, exactly like an inline `Task` inside a group isn't individually addressable. It still composes and runs correctly (each child contributes its context default). If you want `libs.build`, you bind `libs`. I added `test_inline_project_composes_without_namespace` to pin this: the composite resolves and runs, and no `.build` key is mounted.

## Your inline comments + Copilot

| Comment | Resolution |
|---|---|
| `execution.py:16` — why `noqa: TC003`? should be in `TYPE_CHECKING` | You're right — `Path` is annotation-only. Moved into the `TYPE_CHECKING` block, dropped the `noqa`; the one runtime use (a doctest) now does `>>> from pathlib import Path`. |
| `execution.py:77` — docstring on why `base` can be `None` (SSOT) | Added, pointing at `spawn_cwd` as the resolution SSOT. |
| `compose.py:10` — module docstring slop, remove | Gone. The `compose/` package files each carry a single vague one-liner. |
| `compose.py:384` — `_compose_scope` too large; folder + own file | `compose.py` → `compose/` package: `errors.py` (`ProjectLoadError`), `rebase.py` (the `cwd`/`paths`/`when` re-anchoring), `scope.py` (the isolated `_compose_scope` + loaders). |
| `dispatch.py:461` — rename to `load_py_tasks_state`? | Renamed. |
| Copilot `dispatch.py:211` — stale "discovered descendants" | Reworded to "own bindings with its `Project` references resolved". |

## Verification

`camas check` (ruff, mypy, pyright, ty, zuban, pyrefly + doctests) green, `camas coverage` at the enforced 100%, **and** a local `CAMAS_USE_MYPYC=1` wheel passing the full not-slow suite (1243 passed) — the `compose/` package ships interpreted (the non-recursive `main/*.py` glob skips it), the hot modules compiled.

